PR #19, opened 2017-03-28
Remove this file

arcticmouse:
Not sure it is needed, because specificity can be covered in 'IDs and Classes' since there is so much overlap.


mitzirubi:
Just the template

PR #29, opened 2017-04-06
Add to css selector specificity curriculum

mejarc:
* fill in lesson
* fill in practice tasks and challenges
* fill in materials links
* add screenshots
* create sample HTML file


allardbrain:
Hi Melanie, we incorporated this into the revised CSS lesson. Thanks so much!

PR #113, opened 2017-07-26
Add suggested JS libraries, remove some esotery

murphdasmurf:
Removed a bunch of ops asides and threw in some specific (untested by me!) libraries to achieve the OWASP secure development objectives.


allardbrain:
Many thanks, @murphdasmurf! Great additions.

PR #130, opened 2018-01-06
Revision #2 of the Sorting Lesson 

phrenchphry11:
I didn't make a PR for Revision #1.

I did Bubble, Selection, and Merge.  I wasn't sure how in-depth to go.  

My biggest open-ended question:
I think it may be useful to have teacher guidance on QuickSort.  I had students break up into groups to do this, but maybe that's leaving them on their own too much.  

At the very least, I did 3 sorting algorithm lesson plans.  We can probably add more.

### DIFF
--- a/sorting-algorithms/lesson-plan.md
+++ b/sorting-algorithms/lesson-plan.md
@@ -1,73 +1,234 @@
-# Title Here
+# Sorting Algorithms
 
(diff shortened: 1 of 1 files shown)


allardbrain:
This is fantastic! Thank you so much!

PR #136, opened 2018-01-06
Update ajax.md

taygrave:
not totally complete, needs review ofc, might want to consider placing after APIs? not sure


allardbrain:
Nicely done, thank you!

PR #137, opened 2018-01-06
Add some of the "gotchas"

rw:


allardbrain:
Looks great so far, thanks!

PR #150, opened 2018-01-24
Updated and renamed lesson_plan.md for Recursion,RegExp, and Github Collaboration; deleted duplicate lesson plan in Entrepreneurship

chulinguy:


chulinguy:
By the way, the lecture slide calls Git a language; I find this claim to be somewhat misleading because Git is more of a version control system with specific commands to look up/save/modify/delete snapshots of a program.

rcoh:
Thanks!

PR #254, opened 2018-03-10
Add technical assessments.

kenmazaika:
Added technical assessments that can be used as a prerequisite for topic outlines that need a specific technical background.

These assessments will be useful specifically for the topic outlines for `recursion`, but in the future, these assessments should be used to ensure technical capacity for `data structures` specifically for the coding challenges.

In the future, new topics can add assessment types for this.

### DIFF
--- a/technical-assessments/coding-challenges/level-2.md
+++ b/technical-assessments/coding-challenges/level-2.md
@@ -1,0 +1,8 @@
+# Level 2: Coding Challenge Assessments
+
+## Goal
+
+The goal of coding challenge level 2 assessments is to build off of the challenges of Level 1 to solve more and more advanced problems using a programming language.
+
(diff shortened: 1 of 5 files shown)


technical-assessments/coding-challenges/level-2.md
```diff
@@ -0,0 +1,8 @@
+# Level 2: Coding Challenge Assessments
+
+## Goal
+
+The goal of coding challenge level 2 assessments is to build off of the challenges of Level 1 to solve more and more advanced problems using a programming language.
+
+These challenges will involve combining different unrelated coding concepts to solve complex problems.
```
kenmazaika: This would be great to include a bunch of curated CodeWars challenges here.

PR #255, opened 2018-03-10
Add details for Fibonacci

kenmazaika:
Add details for the recursion lesson plan.

### DIFF
--- a/recursion/lesson_plan.md
+++ b/recursion/lesson_plan.md
@@ -13,13 +13,23 @@ About ___ hours
 - JS 4 lesson
 
 ### Motivation
-- Recursion allows programmers to leverage the speed and memory capacity of modern computers to write elegant algorithms
+
+Recursion is a programming concept that allows you to call an original function from within the same function.  This concept is leveraged heavily in _functional programming languages_.
+
+Why is this worth knowing?
+
+* Recursion allows programmers to leverage the speed and memory capacity of modern computers to write elegant algorithms.  This can be leveraged extensively in distributed programming languages like `go`, `erlang` and `elixir` to be remarkably fast on machines with a lot of computing power.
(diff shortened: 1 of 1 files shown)


recursion/lesson_plan.md
```diff
@@ -38,16 +52,48 @@ About ___ hours
 
 ### Things to Remember
 - You can solve all recursion problems with a while loop and vice versa 
-        - While loop solutions usually run faster and take up less computer memory
```
kenmazaika: This isn't true for programming languages with `Tail Call Optimization`.  Since I don't think this should get into tail call optimization in these lessons, I don't think we should include this in the lesson outline.

recursion/lesson_plan.md
```diff
@@ -38,16 +52,48 @@ About ___ hours
...
+
+The value of `10` factorial, for example, can be calculated by:
+
+```
+10 * 9 * 8 * 7 * 6 * 5 * 4 * 3 * 2 * 1
+``` 
+
+> Write a function that uses recursion to calculate the factorial of a number.
+
+#### [Challenge] - Fibonacci
```
kenmazaika: It would be helpful to add a diagram of how to calculate Fibonacci number's here, but volunteering ended before I found a good one!

Karla-Isabel-Sandoval: image and 
image and 
image are some posibilities.

kenmazaika: I think the top two are super helpful!

recursion/lesson_plan.md
rcoh: I think this is probably worth leaving in -- even in languages with TCO, this isn't always the case. In any case, since they're only learning non-TCO languages it seems worth keeping. When later in their careers, the apprentices learn about languages with TCO they can learn about exceptions to this rule.

PR #256, opened 2018-03-10
modify chart rendering code and add some docs

aostiles:


aostiles:
screen shot 2018-03-10 at 2 22 42 pm

rcoh:
thanks for cleaning this up @aostiles !

PR #270, opened 2018-03-14
Update / proof recursion lesson plan

nlacock:
- fixed a few typos
- re-wrote the motivation section to look more like the template
- changed 'independent practice' problem to a new one: factorial was listed as a practice problem twice and also appears in some of the supplementary material.


Karla-Isabel-Sandoval:
@nlacock Way to step it up with the Recursion Topic Outline. Your additions are great! Thank you so much.

PR #277, opened 2018-03-14
Update MongoDB topic outline - Challenge

zzyou:
Changed June back to July. Because if January is 0, then July is 6.


Karla-Isabel-Sandoval:
Great catch @zzyou thanks for updating the TO!

PR #298, opened 2018-03-28
Updated Information to diversity and inclusion section

smcueto:
Added Title and Link to mini slides


smcueto:
* to diversity and inclusion section

PR #300, opened 2018-03-28
Add Runtime Complexity Lecture Video Slides 

zpaterson:
I noticed that the Runtime Complexity Lecture Video Slides were in the apprenticeship schedule, but not in the curriculum, so I thought it would be helpful to add them.


rcoh:
LGTM thanks!

PR #387, opened 2018-05-24
Add sample bash and sql scripts for database setup and teardown

bnak:
Short readme with link to bash tutorial, and sample .sh and .sql scripts for database creation


bnak:
Let me know what you think @MichelleGlauser

bash/create_db_sample.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
```
rcoh: Can you add `set -e` so that errors will cause the script to exit? Probably a good habit  for the apprentices to get into

PR #391, opened 2018-08-13
Database migrations lesson

jeffcompton:
Slides are part of the lesson: (https://docs.google.com/presentation/d/118Fg5gaMPJBQdZV8n9dcAiYLyX5BseVUcJ1fSANdm4E/edit#slide=id.g4032745045_0_3). I'd appreciate feedback on those as well.

### DIFF
--- a/databases/database-migrations.md
+++ b/databases/database-migrations.md
@@ -46,7 +46,10 @@ To ensure that your migration works correctly, it is a good idea to apply it to 
 
 Make sure to not slow down the database for other users during a migration. Migrations on large tables can be slow, and when migrations are in-progress, any other changes to modify the table are blocked until the migration finishes. It is a good idea to apply migrations during a low-traffic time, so it is the least disruptive to users.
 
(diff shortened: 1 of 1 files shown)


databases/database-migrations.md
```diff
@@ -0,0 +1,94 @@
...
+
+To ensure that your migration works correctly, it is a good idea to apply it to a fake dataset first.
+
+Make sure to not slow down the database for other users during a migration. Migrations on large tables can be slow, and when migrations are in-progress, any other changes to modify the table are blocked until the migration finishes. It is a good idea to apply migrations during a low-traffic time, so it is the least disruptive to users.
+
+Include SQL commands in a transaction only when it is detrimental for some commands to succeed and others to fail. Imagine you write two SQL statements to transfer money from one bank account to another. The first statement withdraws money from one account, and the next statement deposits money to another account. They should both be wrapped in a transaction, so that it is impossible for one statement to succeed and the other to fail. Statements that include updating a user's profile picture, and updating a user last active timestamp in a database are pretty unrelated, and shouldn't be wrapped in a transaction. Using transactions when they aren't needed adds unnecessary complexity to your SQL code, and also hurts performance.
```
rcoh: This sentence was a bit confusing to read. Maybe:
```Use transactions only when it's important that separate SQL commands succeed or fail as a unit```

I'd also start by talking about when you'd want transactions in a migration (almost always, and definitely if you're adding or removing columns)

Most transaction running machinery will assume that the individual migrations are running in transactions anyway because otherwise it's impossible to resume from a failed migration.

jeffcompton: @rcoh thanks for the feedback. I updated the section and trimmed it down quite a bit.

It may be a good idea to break out transactions entirely -- for now I'm going to keep the section fairly concise.

rcoh:
Looks great thanks!!

PR #430, opened 2018-09-07
Callbacks and Promises

sherzader:
#402 
Adds Javascript 9 - Callbacks and Promises.


sherzader:
This can be done in conjunction with Lesson 4 or 5 since those topics are lighter.

PR #475, opened 2018-10-26
Cosmetic Improvement on formatting

r0lodex:


alodahl:
nice try @r0lodex!  We do have some really easy issues in our to-do list here; I  suggest the "Strengthen motivation sections" where all you have to do is think of a company that uses that technology and add a few lines, or the "Diversity, Inclusion..." issue that just needs a few slides in a slideshow to be copied into a markdown file:
https://github.com/Techtonica/curriculum/projects/2

PR #486, opened 2018-11-16
Add resource for beginners about functional programming

jdq22:
The link I added is an awesome resource I found a few years back. It's interactive, so it's great for getting immediate results on whether or not your solution is correct.

### DIFF
--- a/functional-programming/FP.md
+++ b/functional-programming/FP.md
@@ -20,7 +20,7 @@
 ### Materials
 
 - [What is Functional Programming](https://medium.com/javascript-scene/master-the-javascript-interview-what-is-functional-programming-7f218c68b3a0)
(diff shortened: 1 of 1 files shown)


jdq22:
There are 41 exercises, and I'd imagine each exercise would take ~15 minutes. This resource is meant to be "go at your own pace." This can be used as an advanced "check for understanding" or as a beginner's exercise.

jdq22:
@alodahl I'll add more info to the file about the resource in terms of "what to expect."

jdq22:
@alodahl Please let me know what you think about the additional copy!

alodahl:
This will be very good for apprentices to know.  Thank you Jenna!!

PR #507, opened 2018-12-29
Trewaters api

Trewaters:
Hi everyone, I talked with @MichelleGlauser and she recommended I make a pull request. 

I was looking over the “API and JSON” part of the curriculum. I saw an opportunity to add a few notes that will hopefully enhance the material your students are covering. Below is an outline of why I made the additions to the current curriculum.

- Status codes are important for working with APIs. A status code is the simplest and most consistent way programmers can verify  they are communicating with the REST API. If there are errors the status code can give clues as to where they should start troubleshooting.
- HTTP methods should be mentioned. GET and POST are used in the examples but not explained. Take a minute to point out they are not the only ones but the ones they will use the most often. Plus they use them every time a web page is viewed.
- Quickly mention REST because it is terminology often used in the software dev environment.
- ALSO a good opportunity to introduce Chrome Dev Tools

### DIFF
--- a/api/apis-and-json.md
+++ b/api/apis-and-json.md
@@ -48,6 +48,8 @@ APIs let you access a wealth of services that other companies and people have wr
     .then(json => console.log(json))
     ```
     Your console should get a response of the same 100-something posts you saw in the browser.  But what if you only wanted one?
+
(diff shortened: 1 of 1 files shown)


alodahl:
What is your expertise? I am trying to get the last few issues finished for the curriculum before the new cohort starts in a week... would you be interested in picking up a to-do issue in our “projects” section?

PR #528, opened 2019-01-15
Update motivation learning to learn

ch-dewayne-perry:


ch-dewayne-perry:
Yes I have discovered that google uses the growth mindset to help their employees and products. Made an edit

Issue #501, opened 2018-12-18
Create a topic on adding enzyme tests to a react project


alodahl:
The ["testing" lesson](https://github.com/aTechtonica/curriculum/blob/master/testing-and-tdd/testing-and-tdd.md) has an overview of how to test applications, but no specifics about including enzyme tests in a node project.

- [x] Write a enzyme test topic outline here: https://github.com/Techtonica/curriculum/blob/master/testing-and-tdd/enzyme-testing.md
- [x] Create an enzyme test slideshow here: https://docs.google.com/presentation/d/1GaoHCdvQ2UplDjapp2wG-VRsxlqtzFqIpOtIqvKNYOo/edit?usp=sharing
- [ ] Find or create a node enzyme test video, and add the link to the materials section in the topic outline file.

The topic trio should include:
- [x] how to label files
- [x] what, if any, packages, and how to include them in test files
- [x] how to run tests and debug in the command line.
- [x] resources for docs and examples
- [x] video
- [x] slideshow should include images of all above processes, either from found video, or screenshot. The only exception is if you find a tutorial that already does this very well.

judytuna:
@alodahl and @amychan331, i wanted to let yoy know that @flyrightsister has donated access to her video on testing in react with Jest and Enzyme! 

public video link to include in curriculum: https://www.udemy.com/react-testing-with-jest-and-enzyme/

private link will be shared with apprentices directly. 

thank you!!!

amychan331:
@judytuna Thanks! I will make sure to include this in. Since the video is about 6 hours, would this be in the material section, not the lesson? I always thought lesson section is activity done in class, and that would make 6 hours too long. @alodahl ?

judytuna:
thank you so much for pointing that out--@alodahl would you agree about putting it i. "supplemental materials" ?

alodahl:
@judytuna @amychan331 In the past, I've had them spend whatever may be left of the hour on the tutorial as the lesson/guided/independent practice, and as the ### Extensions, tell them to finish the rest of they have time. Does that sounds good?  Amy, would you mind adding that to the outline somehow?

PR #556, opened 2019-01-26
Create curriculum topic outline for enzyme

amychan331:
@alodahl @judytuna for #501


alodahl:
This is perfect, thank you so much @amychan331 !

PR #581, opened 2019-02-11
Update web-security.md

bobbylcraig:
Nothing huge. Just some more links and filled empty space in the "Check for Understanding".


bobbylcraig:
@alodahl Closer?

PR #587, opened 2019-02-17
Reframe from "Backend" testing to "Integration" testing

zspencer:
This is an initial reaction commit. Not an "omg u must do this instead" commit. I'm not entirely convinced splitting hairs over unit vs integration testing is worthwhile. I 100% agree that backend testing and front end testing are different skills with different techniques. (front end is much more event-driven, backend is often more promise driven).

I'm hopeful that by disambiguating between _integration_ and _unit_ testing we provide a box
 for students to put information in as their mental model refines.

LMK if you think it's worth continuing down this path in this PR, or if you'd prefer we don't make such a significant change!

### DIFF
--- a/testing-and-tdd/integration-testing.md
+++ b/testing-and-tdd/integration-testing.md
@@ -23,18 +23,36 @@ Here are links to lessons that should be completed before this lesson:
 
 ### Motivation
 
-Up to now we've been talking about unit testing within the context of making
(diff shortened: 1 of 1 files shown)


alodahl:
@zspencer You should go for it, as long as you define the difference in the outline or slides. 👍🏻

PR #588, opened 2019-02-19
Updating to include tie in to logout material on slides

ch-patrickfoubert:
Admittedly it didn't feel like there was a good place to put logout, since the instructions / pairing up work ends at verifying a fix to configuring for Google login. I figure since this is really the goal of the session, I moved the challenges bit to be after the working session, and in there included a challenge based on logout.

- Moved the challenge section to be after "Check your Understanding"
- Added an additional challenge to tie into the logout slide added for Auth0.


alodahl:
@ch-patrickfoubert Thank you!  Have you changed the slideshow yet?

Issue #560, opened 2019-01-31
Ongoing:  Add a suggestion to our Women in Tech Resource List!


alodahl:
- Help our apprentices find support and resources by adding a suggestion to [this topic outline about Resources for Women in Tech](https://github.com/Techtonica/curriculum/blob/master/women-in-tech-resources/resources-list.md). 

For FAQs on contributing, see https://github.com/Techtonica/curriculum#volunteering

rudri182:
@alodahl I would like to work on this issue under GSSOC'19

alodahl:
Thank you!  Submit it on March 1 and i will count your PR for GSSoC!

thetaniyagupta:
I would also like to contribute to this issue under GSSOC'19

Lakshita7:
I would like to work on this issue under GGSOC'19

alodahl:
@thetaniyagupta @Lakshita7 👍. Go ahead and make a PR on March 1 😊

alodahl:
@thetaniyagupta @Lakshita7, since we are a US Non-profit, please be sure to fill out this form before you start this project: https://docs.google.com/forms/d/e/1FAIpQLSeW0mo-Dpsig70374UEPvzexpas-31Ost_HsFwm0kjNOxtbtg/viewform?c=0&w=1

ronika-das:
would like to contribute to this issue. Do the resources/meetups etc have to be strictly for women/girls only? there are a couple of them open for all

alodahl:
@ronika-das , if you think it's a resource that would help women and non-binary people network, you can add it.

I emailed you an invitation to contribute; once you accept that, you can assign yourself to this issue.

devakisj1:
hello ma'am , i would like to work on this issue too.

alodahl:
@devakisj1 I emailed you an invitation to contribute; once you accept that, you can start working on the issue and I will assign you.

Soumya0803:
@alodahl  Ma'am,I would like to contribute to this issue

Manvityagi:
@alodahl I want to work on it as my first issue? Please guide.

himadrinayak:
@alodahl I would like to work on this issue under GSSOC'19

vaishali614:
@alodahl i would like to work on this issue under gssoc'19

vsharma20v:
@alodahl I would like to add more groups to this.

Rita-Okonkwo:
[...]

PR #628, opened 2019-03-04
added more resources

devakisj1:
#560 added more resources


manaswinidas:
Please mention the issue that you are solving. It will be easier for us to give scores😃

PR #636, opened 2019-03-06
Add instructions for adding react frontend to eventonica

lalpert:


Kundan28:
@lalpert The line 38 and 39 are blank so one of them could be removed.

PR #679, opened 2019-03-19
Update data-structures-algorithms-sprint.md

tiffanyjstuart:
Removed duplicate tree link (it's already included later in the list)

This isn't associated with any issues.


tiffanyjstuart:
@alodahl I noticed this minor issue when looking at the schedule for today.

Issue #594, opened 2019-02-23
Augment OOP outline


judytuna:
Make the following changes in this outline: 
https://github.com/Techtonica/curriculum/blob/master/javascript/javascript-7-oop.md

## Write down a list of terms. 

People can find resources on their own.

## Generate "Check for Understanding" questions for worksheet. 

For discussion in small groups

## Add to Supplemental Resources 

Polymorphism

https://dev.to/jvanbruegge/what-the-heck-is-polymorphism-nmh

nodeschool?

alodahl:
@judytuna, can you explain your nodeschool comment in the description?

imskr:
I would like to work on it!  And tell me about nodeschool?

alodahl:
sure, @imskr .  Hopefully Judy will tell us more about NodeSchool before you finish.

vaishali614:
@alodahl i would like to work on it! can you assign me this issue?

PR #688, opened 2019-03-21
Update javascript-7-oop.md

imskr:
Issue #594

### DIFF
--- a/javascript/javascript-7-oop.md
+++ b/javascript/javascript-7-oop.md
@@ -29,6 +29,7 @@ Virtually all companies will use object-oriented programming in their code in or
 **Participants will be able to:**
 - Understand some of the main concepts of object-oriented programming
 - Start developing an understanding of objects and classes
+- Find resources on their own
 
 ### Specific Things To Teach
 - Classes and objects
@@ -335,4 +336,6 @@ Explain your completed Codecademy lesson code to a partner.
 - [Reference: JS Object Basics - Mozilla Developer Network](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Objects/Basics)
 
 - [Chapter: Object-Oriented Programming - Eloquent JavaScript](http://eloquentjavascript.net/1st_edition/chapter8.html);
+
+- [What the heck is Polymorphism? by Jan van Brugge](https://dev.to/jvanbruegge/what-the-heck-is-polymorphism-nmh)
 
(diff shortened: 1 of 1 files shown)


imskr:
I've added extra resources too. @alodahl Please review.

imskr:
@Shwetha-Acharya Thanks for reviewing, I will remove it.

imskr:
Ok got it. I will add those terms.

imskr:
@alodahl Done.

javascript/javascript-7-oop.md
```diff
@@ -321,8 +321,19 @@ The `Loan` constructor is using the `calculateDueDate()` method polymorphically.
...
-Work through [this Introduction to Classes on CodeAcademy](https://www.codecademy.com/courses/learn-javascript-classes/lessons/classes/exercises/introduction?course_redirect=introduction-to-javascript).
+* Work through [this interactive lesson on Advanced JavaScript Objects](https://www.codecademy.com/courses/introduction-to-javascript/lessons/advanced-objects/exercises/adv-intro?action=resume_content_item).
+
+* Work through [this Introduction to Classes on CodeAcademy](https://www.codecademy.com/courses/learn-javascript-
+classes/lessons/classes/exercises/introduction?course_redirect=introduction-to-javascript).
+
+* Find definition of following terms on your own:
+   * Namespace
+   * Hoisting
+   * Arity
```
Shwetha-Acharya: Just curious, are you trying to establish the fact that the arity property which was used to return the number of arguments expected by the function, no longer exists and has been replaced by the Function.prototype.length property?? @alodahl What do you think, what is good to mention here, arity or Function.prototype.length?

Shwetha-Acharya: I would say to replace arity by Function.prototype.length.

Shwetha-Acharya: Anyways as a concept/definition it is well and good to explore about arity. But you should establish the fact specific to javascript. You can suggest to go through the link: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/arity after mentioning Arity.

imskr: Thank You for pointing out. Arity is now replaced by Function.prototype.length but Arity can also be used. I will add it too.

alodahl:
Tip for future: describe specific changes in PR rather than * Update javascript-7-oop.md.  Great work, thank you =)

Issue #730, opened 2019-04-08
Make sure Prerequisite sections are correctly linked (1/2)


alodahl:
## Task:

This is a single-contributor issue, and should all be in one PR.

Check the first half of the outlines in the [ReadMe Table of Contents](https://github.com/Techtonica/curriculum/blob/master/README.md#table-of-contents ) and ensure that the prerequisites are linked correctly. 

 **For example, if you see this:**
### Prerequisites
- Command Line Lesson

**You should change it to this:**

### Prerequisites
- [Command Line Lesson](/command-line/command-line-interface.md)

**Complete the Prerequisite sections in these sections’ outlines:**
- [ ] Onboarding 
- [ ] Programmer Tools
- [ ] Client-side programming

Most of them should already be done, but your job is to track down the last ones.

Exceptions: if the prerequisite is obviously advice rather than a specific lesson, you don’t have to link it. However, since you’ll be exploring our curriculum anyway, use critical thinking to add an unmentioned prerequisite if it seems like it should be there.

tiffanyjstuart:
@alodahl In https://github.com/Techtonica/curriculum/blob/master/learning-to-learn/learning-to-learn.md link should "Expectations While at Techtonica" lesson link to the google doc with the "Apprentices Expectations"?

tiffanyjstuart:
Completed all the Onboarding Prerequisite sections.

tiffanyjstuart:
@alodahl Some of the lessons in the Prerequisite sections are missing from the ReadMe Table of Contents (https://docs.google.com/document/d/1IWny-IXB-XDWW21l3JODwPuh1QOVBFQbii_DjvF0g00/edit?usp=sharing, https://github.com/Techtonica/curriculum/blob/master/conflict-resolution/conflict-resolution.md, https://github.com/Techtonica/curriculum/blob/master/diversity-inclusion-bias/inclusion.md)

alodahl:
@tiffanyjstuart great catch! Would you have time to add them based on the order in the Roadmap, or should I make an issue for that?

yati1998:
@alodahl  can I work on this?

tiffanyjstuart:
@alodahl No need to make an issue, I made a new PR (https://github.com/Techtonica/curriculum/pull/742) adding them to the Roadmap and consolidating my other PRs

yati1998:
@alodahl do I need to add prerequisites to the links in which none is already mentioned?

alodahl:
[...]

PR #743, opened 2019-04-11
Issue #730

nchazin:
Update prerequisite links on the javascript lessons  - I realize there is more to do here, but this seemed to be a good chunk to update.


alodahl:
Hi @nchazin, thanks for keeping an eye out.  Do you think that idea is covered by this PR, or do I need to add something?  https://github.com/Techtonica/curriculum/issues/730

PR #763, opened 2019-04-15
updated prerequisites section

yati1998:


yati1998:
@alodahl which HTML/CSS workshop is mentioned here? I cannot find any such. Are these workshops the same as the lesson..?

PR #775, opened 2019-04-19
Insert link in languages section

rudri182:
Link for JS and SQLQuery is inserted in languages section for eventonica-project.md


alodahl:
this should not be in a separate PR.

PR #913, opened 2019-06-29
converted checks for understanding q's -- chrome dev tools

marieatterbury:
Converted checks for understanding into small-group questions.


alodahl:
Thanks for your changes, Marie.  I made the review about the rest of the document into another issue.

Issue #910, opened 2019-06-23
Fix Typos in Event Recommender guidelines


alodahl:
Fix typos pointed out in these reviews:
https://github.com/Techtonica/curriculum/pull/906

PR #914, opened 2019-06-30
Fix Typos in Event Recommender guidelines

vvscode:
https://github.com/Techtonica/curriculum/issues/910

### DIFF
--- a/javascript/event-recommender-part2.md
+++ b/javascript/event-recommender-part2.md
@@ -27,9 +27,9 @@ Now that you've finished Part 1 and your Event Recommender class is complete, yo
 
 ### UI Instructions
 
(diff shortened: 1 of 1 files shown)


vvscode:
Sure. Done

vvscode:
@alodahl  the rest are related to code in gist - I'm not sure if forking that gist to make changes is the right approach. But if you assume that should be done - I'm ready

alodahl:
Maybe I mis-titled the issue, but yes, that would be great!

vvscode:
@alodahl  Gist is updated

vvscode:
@alodahl  Actually I see no sense to use gist for such cases as well. Placing code embedded to the document makes everything much more accurate (imho)

alodahl:
Would you mind embedding it, then? I didn’t realize that the gist is now on your account.

Issue #700, opened 2019-03-24
convert check for understandings into questions (multiple assignees encouraged)


alodahl:
Related to #575 
## Task:

 Change the linked outlines below so that your PR ensures:

Look through topic outlines, and change **any 2** so that your PR ensures:
- It’s Check for Understanding section tells apprentices to form small groups and answer at least 2 long-answer or more shorter-answer questions so they can self-assess whether they understand the material.
- The questions should check whether they’ve achieved **every** part of the “objectives” section at the beginning of the outline.
- The updated CFU section should not change the total “projected time” dramatically. If the “projected time” changes as a result, be sure to update it.
- Don’t just delete what’s already in CFU. (See next points)
- if the CFU section already contains something valuable, we can consider putting it in the independent practice or challenge section instead. Don’t forget to adjust projected time.
- if the CFU section contains something less valuable, you can move it to the supplemental resources section.
- if you finish, or notice a Check for Understanding already looks fine, please check it off on this list or let a Techtonica admin know it's complete
- A single PR can be made for two edited Check for Understanding sections, earning you a Hacktoberfest contribution.

* Here is a great example of what we want for our Check for Understanding sections from [our TDD outline](https://github.com/Techtonica/curriculum/blob/master/testing-and-tdd/testing-and-tdd.md):
### Check for Understanding

Form small groups and discuss:
- What is an automated test?
- What is TDD?
- What are some advantages of TDD?
- What are some disadvantages of TDD?
- Where should your tests live in your project?
- How do you run your tests?

——————————-

Topic outlines whose materials sections not been reviewed:
- [x] https://github.com/Techtonica/curriculum/blob/master/personal-brand/personal-branding.md
- [x] https://github.com/Techtonica/curriculum/blob/master/professionalism/professionalism.md
- [x] https://github.com/Techtonica/curriculum/blob/master/ergonomics/ergonomics.md
- [x] https://github.com/Techtonica/curriculum/blob/master/learning-to-learn/learning-to-learn.md
- [x] https://github.com/Techtonica/curriculum/blob/master/roles-in-tech/roles-in-tech.md
- [x] https://github.com/Techtonica/curriculum/blob/master/welcome-tips-rules/tech-industry-tips.md

- [x] https://github.com/Techtonica/curriculum/blob/master/command-line/command-line-interface.md
- [x] https://github.com/Techtonica/curriculum/blob/master/command-line/vim.md 
[...]

PR #954, opened 2019-09-02
Sniemar gitignore

sniemar:
Fixed the issue #700

### DIFF
--- a/git-version-control/gitignore.md
+++ b/git-version-control/gitignore.md
@@ -118,8 +118,10 @@ You should no longer see *.my-secret-keys* in your untracked files, because git 
 
 ### Check for Understanding
 
-- Small group discussion:
+Form small groups and discuss:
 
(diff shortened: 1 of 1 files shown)


git-version-control/gitignore.md
```diff
@@ -118,8 +118,10 @@ You should no longer see *.my-secret-keys* in your untracked files, because git
 
 ### Check for Understanding
 
-- Small group discussion:
+Form small groups and discuss:
 
   - What sorts of files should you add to your .gitignore?
-
+  - How can you add to .gitignore each file of one type?
```
alodahl: Please change to: how can you add all files of one type to your .gitignore file?

alodahl: the objectives they should be making sure they understand are: 
- [x] Adding unwanted files to their .gitignore
- [ ] Know what data to protect from public view
- [ ] Know which files don't need to be committed

PR #963, opened 2019-09-16
Added `Check For Understanding` questions to VSCode and Asking Good Questions lessons

### LINKED ISSUE #905: convert check for understandings into questions (Part 2)
Related to #575 and #700 

- [ ] https://github.com/Techtonica/curriculum/blob/master/asking-good-questions/asking-good-questions.md
- [ ] https://github.com/Techtonica/curriculum/blob/master/vs-code/vscode.md

## Task:
Change the above topic outlines so that:
- Their Check for Understanding sections tell apprentices to form small groups and answer at least 2 long-answer or more shorter-answer questions so they can self-assess whether they understand the material.
- The questions should check whether they’ve achieved **every** part of the “objectives” section at the beginning of the outline.
- The updated CFU section should not change the total “projected time” dramatically. If the “projected time” changes as a result, be sure to update it.
- Don’t just delete what’s already in CFU. (See next points)
- if the CFU section already contains something valuable, we can consider putting it in the independent practice or challenge section instead. Don’t forget to adjust projected time.
- if the CFU section contains something less valuable, you can move it to the supplemental resources section.
- if you finish, or notice a Check for Understanding already looks fine, please check it off on this list or let a Techtonica admin know it's complete
- A PR can be made for one edited Check for Understanding section, and each will earn you “beginner” points if you are part of gssoc.

* Here is a great example of what we want for our Check for Understanding sections from [our TDD outline](https://github.com/Techtonica/curriculum/blob/master/testing-and-tdd/testing-and-tdd.md):
### Check for Understanding

Form small groups and discuss:
- What is an automated test?
- What is TDD?
- What are some advantages of TDD?
- What are some disadvantages of TDD?
- Where should your tests live in your project?
- How do you run your tests?

shubha-rajan:
fixes #905

### DIFF
--- a/vs-code/vscode.md
+++ b/vs-code/vscode.md
@@ -73,7 +73,7 @@ Techtonica staff will assign pairs.
 
 1. Follow the instructions in the slides for downloading and installing VS Code if it is not already downloaded to your machine.
 
-2. Using [VS Code's Getting Started manual](https://code.visualstudio.com/docs) or other online resources, discover how to:
+1. Using [VS Code's Getting Started manual](https://code.visualstudio.com/docs) or other online resources, discover how to:
 - Find and replace multiple occurrences of one word in the same file
 - Click on multiple characters in one file at the same time
 - Comment and uncomment text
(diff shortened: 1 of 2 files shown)


vs-code/vscode.md
```diff
@@ -91,18 +101,13 @@ Briefly summarize your findings so you can share with the class.
...
-- Open a specific folder in VS Code
-- Open a specific file in VS Code
-- Find and replace multiple occurrences of one word in the same file
-- Click on multiple characters in one file at the same time
-- Comment and uncomment text
-- Indent and dedent lines of text
-- Move one line up or down
-- Open a file side by side
-- Open integrated terminal
+1. How do you open a specific file or folder in VS Code?
```
vvscode: actually it's better to use list in 1-numbered style

```
1. One
1. Two
1. Three
```

it gives

1. One
1. Two
1. Three

and in case of editing no need to renumber items

shubha-rajan: I changed these to bullets for consistency but changed other ordered lists in the file to 1-numbered style.

alodahl:
Thanks so much Shubha, this is perfect! And thanks @vvscode for reviewing.

PR #964, opened 2019-09-19
Connection method update and bug fixes in Eventonica project

### LINKED ISSUE #958: Replace deprecated 'connect' function in API project with updated version
Recent curriculum feedback: Connection function in [Eventonica](https://github.com/Techtonica/curriculum/blob/master/projects/eventonica-project.md) is deprecated.

Replace connection function found in the linked file above with a current working version. This may require a change to the [starter code](https://github.com/Techtonica/curriculum/tree/master/projects/Eventful-API-test-page), the [project directions](https://github.com/Techtonica/curriculum/blob/master/projects/eventonica-project.md), or both.  After you've tested that it runs correctly, make a pull request with the updated file.

For FAQs on how to contribute, see: https://github.com/Techtonica/curriculum#volunteering

shubha-rajan:
Fixes #958 

I replaced the deprecated connection method to use `pg.Pool` and pass a `pg.Client` into `app.startQuestion` (see https://node-postgres.com/api/pool). 

I also went ahead and fixed some other bugs I found, like a typo and a missing script in package.json, and clarified the process of requesting API keys in the project direction.

### DIFF
--- a/projects/Eventful-API-test-page/index.js
+++ b/projects/Eventful-API-test-page/index.js
@@ -1,17 +1,20 @@
-const connection = require('./connection');
+const pool = require('./connection');
 const app = require('./app');
 
- const pgConnect = () => {
-  connection.connect((err) =>{
+const pgConnect = () => {
+  pool.connect((err, client) => {
     if (err) throw err;
 
-     console.log('Welcome to Eventonica')
+    console.log('Welcome to Eventonica')
     console.log("connected as Administrator");
 
-     app.startQuestion(()=>{ connection.end() });
+    app.startQuestion(() => {
+      client.release();
+      pool.end().then(() => process.exit());
+    }, client);
   })
 }
 
- // *Uncomment below line once you have mySQL setup
+// *Uncomment below line once you have postgreSQL setup
 
- pgConnect();
+pgConnect();
--- a/projects/Eventful-API-test-page/app.js
+++ b/projects/Eventful-API-test-page/app.js
@@ -1,10 +1,9 @@
 const inquirer = require('inquirer');
-//connection available to all
-const connection = require('./connection');
 
- const app = {};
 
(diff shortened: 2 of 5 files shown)


projects/Eventful-API-test-page/index.js
```diff
@@ -1,17 +1,20 @@
...
+  pool.connect((err, client) => {
     if (err) throw err;
 
-     console.log('Welcome to Eventonica')
+    console.log('Welcome to Eventonica')
     console.log("connected as Administrator");
 
-     app.startQuestion(()=>{ connection.end() });
+    app.startQuestion(() => {
+      client.release();
```
gsong: I don't think using callback for `pool.connect` requires you to manually release the client. It's also not clear to me why we're checking out a client here from the pool, since we don't do anything with it.

gsong: Oh, is this supposed to be a template of sorts for the students to follow?

shubha-rajan: This is starter code for the project, so yeah, the intention is for it to be a template. Ideally, students would check out a client and release it each time they query the db in app.js, but to get the project running they should successfully connect to the database to know it's working.  But you are correct that using a callback automatically releases the client.

shubha-rajan: Alternatively, `connection.js` could create a `Client` instead of a `Pool`, and students could just use the one client in all of their queries. Although the node-postgres docs recommend using pools instead of individual clients

gsong: No, I think using a pool is good practice.

gsong:
@shubha-rajan This looks good from my perspective. I don't know what the merge procedure is here, so I'll let you push that through.

shubha-rajan:
I'm not sure if I should merge myself or wait for @alodahl to merge

PR #973, opened 2019-10-01
Added roles in agile and MVP

sophi-li:
Added to "Specific things to teach"


alodahl:
@sophi-li Thanks for your help!  I was able to add these issues:
https://github.com/Techtonica/curriculum/issues/977
https://github.com/Techtonica/curriculum/issues/976

PR #974, opened 2019-10-01
Proof reading assessment objectives

Eddenm:
Just general proof reading on some of the assessment objectives - to make slightly more user friendly.


Eddenm:
Yes - I can. Will look through these for you.

PR #985, opened 2019-10-06
Added instructions to switch to csv mode in sqlite3

julinvictus:
```.mode csv``` was missing from original instructions. I spent time trying to import those csv files until I found instructions at google: https://www.sqlitetutorial.net/sqlite-import-csv/


julinvictus:
That should be enough to make it easier.

PR #1037, opened 2019-10-15
Update http-request-practice.md

### LINKED ISSUE #616: Write up instructions for "HTTP request field day" exercise based on existing notes
Write a topic outline [here](
https://github.com/Techtonica/curriculum/blob/master/api/http-request-practice.md) based on the following notes:

Do the same two requests over and over again six different ways! Why? Compare and contrast: how are they different? How are they similar? 

1. Postman (a Mac app that’s a GUI for curl) (this should be trivial as they've been using it all week. start with something easy on purpose.)
  - at the end of the exercise, have them discover in the Postman GUI that there's a menu option to grab the request in `curl` syntax. what do we learn about Postman? it's probably a GUI for curl! 
2. `curl` on the command line, in their terminal
3. `XMLHttpRequest` in raw javascript in a javascript console in a browser
  - to avoid CORS issues, don't use about:blank. instead, just hit `localhost:3000/events` in your browser and open a javascript console there. boom, same origin! =D
  - bonus reading about CORS: https://medium.com/netscape/hacking-it-out-when-cors-wont-let-you-be-great-35f6206cc646 (we aren't encouraging people to use a proxy at this time though. just read this for info) 
4. window.fetch in that javascript console (just to see that it returns a promise)
  - explain that this is newer javascript
5. http.request in node (it comes with node), in the node repl
6. node-fetch in node, again in the node repl but after `npm install --save node-fetch` in the client project they're about to work on

Apprentices will look up the syntax (so this is also an exercise in searching for resources and separating signal from noise), with some hints, like "we told you this is an NPM module. where should the very first place you look for it be? answer: docs on npm, which will link to its github repo. example: https://www.npmjs.com/package/node-fetch and https://github.com/bitinn/node-fetch "

More discussion questions:
- What affordances to higher-level libraries give us? What does "higher-level library" mean? 

A few "Additional Resources" that compare ways to make requests:
- Axios, Request (different from `http.request`), and `http` (the Node standard library): https://flaviocopes.com/node-http-post/ 
- old and long comment thread including examples with `http.request`, Request, Unirest, superagent, node-fetch, axios, `http` using `stringbuilder`, and more. Somehow convey that here's an example of "there are a million ways to do it," so it might take you some time and research to figure out the best way to do something for the current situation, and that's normal. 
- `http.request` resources
  - http://shiya.io/send-http-requests-to-a-server-with-node-js/
  - https://itnext.io/how-to-handle-the-post-request-body-in-node-js-without-using-a-framework-cd2038b93190

img_4573 smaller

For FAQs on how to contribute, see: https://github.com/Techtonica/curriculum#volunteering

AnthonyJr:
fixes #616

### DIFF
--- a/api/http-request-practice.md
+++ b/api/http-request-practice.md
@@ -3,7 +3,7 @@
 ### Projected Time
 
 Example: 30-45 minutes
-- Lesson: ? min
+- Lesson: ? min(s)
 - Guided Practice: ?-? min
 - Independent Practice: ? min
 - Check for Understanding: ?-? min


alodahl:
Hi @AnthonyJr , sorry, but this PR doesn't add a lot of value.  Please consider picking up an easy issue [here](https://github.com/Techtonica/curriculum/projects)!

PR #1041, opened 2019-10-16
Update Check for Understanding Section

elaguerta:
Direct participants to form small groups.
Edit questions to address Objectives, including discussing a hypothetical situation.
Add 20 minutes to total lesson time.


elaguerta:
@alodahl Thanks for reviewing! I added a question asking what participants could do to improve their use of credit cards and their credit score.

PR #1044, opened 2019-10-17
updated check for understanding

jay1723:
Updated the check for understanding for the negotiation page.


jay1723:
Completes issue #1027

PR #1056, opened 2019-10-23
Add android and iOS devloper Roadmap

MatheusUehara:
Adding Roadmap to the study path will help developers in their study by showing what they need to know about each platform.


MatheusUehara:
Yes, i have one suggestion! 

I think we should split mobile development in each technology(Android, iOS, React Native) it will turn the paths more short and consise. 

The split can be in separated MD files...

I can send more PRs with that changes :)

PR #1069, opened 2019-10-27
edited operators and function

tessythomas123:


python/python.md
```diff
@@ -253,6 +260,8 @@ print 'always executed'
 ### Independent Practice
 
 - Create a function that will take a string argument and return that string in all uppercase letters.
+	def uppercase(str):
```
alodahl: This is something the learner is supposed to solve - don’t give them the answer!

python/python.md
```diff
@@ -111,12 +111,19 @@ def add_one(num):
 
 #### Operators
 - Arithmetic:
-	- +, -, *, /, and % (modulus)
+	- +, -, *, /,//(floor division),**(exponentiation), and % (modulus)
```
alodahl: 👏🏻

Issue #1067, opened 2019-10-27
Add example links and explanations to Roles in Tech:  Technical Writing


alodahl:
Add the following links and explanations, reworded if necessary, to appropriate sections in our [Roles In Tech: Technical Writing outline](https://github.com/Techtonica/curriculum/blob/master/roles-in-tech/technical-writing.md).

_________________
Here are some technical writing examples.
Process discussions: https://www.prismnet.com/~hcexres/textbook/proc.html
API documentation: https://nordicapis.com/5-examples-of-excellent-api-documentation/
Walkthrough documentation, perhaps the most overall useful for people who are technically capable but unfamiliar. https://medium.com/product-labs/how-to-write-a-useful-scenario-walkthrough-f48bf40b1b69 
...and here's a good high-level overview of the kinds of documentation technical writers typically get involved with:

https://clickhelp.com/clickhelp-technical-writing-blog/types-of-technical-documentation/

In addition to the above, my team edits a lot of external blogposts too. You can see are work here, polishing other people's articles:

https://engineering.indeedblog.com/blog/
https://medium.com/indeed-engineering

....and occasionally getting to write our own. : ) 

https://medium.com/indeed-engineering/the-evolving-language-of-data-science-89830c5aa0c7

carmelag:
Hi! @alodahl I would like to help here. 
So what is needed to solve this issue, is to add in _technical-writing.md_ file those links along with a good explanation, is it correct?

carmelag:
Hi @alodahl , I have just created the PR related to this issue.
I added the links that were directly related to tech writing but I didn't add the last 3 links (indeed ones). They seem to be less specific than the others.
If it's needed I will add them, but it would be helpful if someone could add more details to better understand what kind of description/headline we could put for them. Thanks

PR #1093, opened 2019-11-05
Add some links

carmelag:
Add some links about technical writing to the Supplemental Materials section. Related issue: #1067 1067


carmelag:
Hi @alodahl 
sorry for the delayed reply, but I was on holiday. Could you please have a look at the new version of supplemental materials section? I committed new changes to the same PR
Thanks

PR #1101, opened 2019-12-18
draft tech-industry-tips improvements about distractions/deep work

vegetabill:
- TODO: find free distraction-control app
  - TODO: find a anonymous or public record example of someone getting fired for private email or IM's so it's a stickier story


vegetabill:
Does anyone have a good distraction-prevention app or Chrome extension?

The part about email I think could be

PR #1104, opened 2019-12-20
[mongo] update lesson guide due to recent changes where mongo is now …

vegetabill:
…a homebrew tap and has a new name

* verified these new instructions this morning on catalina
* removed all the custom instructions for data dir since they are not needed if you use brew service


vegetabill:
I just followed the official guide and was able to connect with the client and do all of the examples for the rest of the lesson so it should be good.

You can probably still run it manually like in the existing version of this tutorial but since this listen isn't about daemons/services it probably makes sense to do it the simplest way.

vegetabill:
@alodahl good idea! I didn't think of that. I checked and didn't see any other references where it would matter. There is a Mongo assessment which doesn't list the commands, so this should be safe to merge.

Created #1108 to track the change.

Issue #1086, opened 2019-11-01
Review Data Visualization Lesson


lalpert:
For the data visualization lesson: https://github.com/Techtonica/curriculum/blob/master/data-visualization/data-visualization.md

Apprentices mentioned this lesson was hard to follow and get working. 

Please do the full lesson and flag any instructions that are confusing, out-of-date, or have errors. 

You can make a PR to fix it or if you don't have time, comment with a list of improvements needed.

sdspikes:
There's a link to https://bost.ocks.org/mike/bar/ in the materials section, then the Guided Practice session uses the code from that page but doesn't mention that it's the source, and both sorta jump in without much context.  The first instruction in the guided practice is "Create a div with 'chart' class, <div class="chart"></div>, in the body of the webpage", -- which webpage?  I think you just want them to make a new/empty html document to use for this, but I was confused for a bit about whether there was a particular existing page I was supposed to be editing (e.g. from the linked tutorial) vs starting a new one.

So, I went through the link.  I don't find it particularly engaging (it doesn't ask me to think for myself at all, and gives me code without motivating why we want to do each step -- it feels more like the author wrote some code to do what they wanted, then explained it line by line rather than thinking like a beginner and helping the reader understand how to go from idea to implementation). 

I also find the structure of it kinda confusing, and hard to follow -- in the "Coding a Chart, Automatically" section, it says "starting with an empty page that contains only a div of class "chart"" but actually wants you to keep the styling as well (and load d3, which the page assumes you know how to do at the top, but never reminds you to actually load it).  It then gives a block of code without explaining it, then goes line by line using a different version of the code where it's all broken out into individual variables rather than chained.  And as mentioned, it uses an old version of d3, which leads to problems with `scale.linear` vs `scaleLinear`, as well as using old style `var`.

Recommendation: fully re-write the linked tutorial with d3 v4 in a way that is more straightforward (in particular, more clarity around what to include where, when it's on a tangent vs asking you to actually modify code, etc), and have it lead to the code that's presented in the current "Let's Make a Bar Chart" section of the lesson (right now there's extra stuff with colors added in without comment, making the code look more daunting than it needs to be, build that in to the tutorial).

The practice bit to get the number at the bottom rather than the top of the bar after the transform was useful, and helped drive home the fact that attrs can be either a value or a function.  I started looking at the pie chart thing, but don't have much more time right now so won't go into that for now.

PR #1107, opened 2019-12-20
Update data-visualization.md

sdspikes:
Bridged more directly from the linked walkthrough, and took it step-by-step from there instead of jumping to a much more complex version.

This is in response to https://github.com/Techtonica/curriculum/issues/1086

### DIFF
--- a/data-visualization/data-visualization.md
+++ b/data-visualization/data-visualization.md
@@ -43,6 +43,7 @@ Companies like [Fidelity](https://www.fidelity.com/) and [Strava](https://www.st
 ### Materials
 - [Data Visualization Slides](https://docs.google.com/presentation/d/1Vzx0eyKt0jNQq70pnAeDQ5X1qOPdjIzCgujSp6dd_1U/edit?usp=sharing)
 - [Let's Make a Bar Chart (20min read)](https://bost.ocks.org/mike/bar/)
+  - Note: uses version 3 of D3, so if you try to follow along as written, you will either need to import v3, or modify the code as in the guided practice below
 
 ### Lesson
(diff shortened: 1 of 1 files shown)


sdspikes:
I considered converting the functions to ES6 lambdas, but not sure if that's something they'd already be familiar with, or what the style standards are for curriculum.

sdspikes:
> This is such a great revision of this outline, thank you!! Just one question... do you approve of the Check For Understanding questions? Would you change anything?

Going through them one by one:

> Discuss how one would go about updating the DOM and utilizing D3.js. What are the general steps to set up a new D3.js visualization?

This seems fairly reasonable, but looking at it as a student I'm not totally sure how I'd answer this.  Are we looking for things like importing it, specific methods that need to be called (e.g. `select`), or a bigger-picture explanation?

> What factors do you consider when selecting a data set?

I feel like the lesson doesn't really cover this -- we ask them to select a data set in the challenge, but don't give any guidance anywhere about it.  If this is something we want them to be able to answer, the lesson should probably include something where we give concrete data sets and talk about tradeoffs of them.  I also don't feel like I know how to answer this question without context -- selecting a data set to do what?  Are we talking about choosing between different sources of data, different formats, different types of info?  

> What do you consider when determining how to display data in a compelling way?

This question makes more sense to me, but I don't think the lesson helps much with answering it.  I guess they're supposed to go through some of the supplemental materials to figure this out.


Takeaway: I think the second question is the weakest/least useful.  I am not sure what I'd replace it with off the top of my head, or what I'd change about the lesson if we do want it in.  I'll think about it a bit and see if anything comes up.

alodahl:
Would you be open to re-writing them to fit the outline as it is now?

Issue #969, opened 2019-09-26
OAuth Guided Practice is incomplete


MichelleGlauser:
From @ngriffin23: "Creating a log out button is incomplete and they also mention to add some code into an Auth.js file but that file is not provided when you download the sample folder."

Resolve this problem in our auth0 outline here:https://github.com/Techtonica/curriculum/blob/master/oauth/o-auth.md

lemon-yellow:
I can take a look at this one!

lemon-yellow:
@alodahl - the sample React app from Auth0 comes with a functioning Logout button - was that not the case in the past? 

Here is the code for the Logout button - https://github.com/auth0-samples/auth0-react-samples/blob/master/01-Login/src/components/NavBar.js#L85-L91

And the logout logic is not in the sample code anymore, it is imported from Auth0's library here- https://github.com/auth0/auth0-spa-js

Does it make sense to have lesson cover another task instead?

PR #1150, opened 2020-01-19
Remove step 4 - logout button

lemon-yellow:
The latest Auth0 sample code for React comes with a functioning
logout button. The files referenced in the instructions no longer exist
and have been replaced with an external Auth0 package.

Addresses https://github.com/Techtonica/curriculum/issues/969


lemon-yellow:
There is already a logout button in the Auth0 sample code that works. But I could update Step 4 to ask the reader to note the logout button after they log in. They can also explore the sample code to find the logout button logic and the logout logic.

lemon-yellow:
@alodahl - I updated Step 4 to ask the reader to review the source code to find the login/logout logic. Let me know if this looks good to you.

Issue #967, opened 2019-09-22
Add new Mongoose lesson link to curriculum records 


alodahl:
Add new Mongoose lesson link (see  #966) to most current Roadmap, the [ReadMe](https://github.com/Techtonica/curriculum#table-of-contents), and that week's [assessment objectives](https://github.com/Techtonica/curriculum/tree/master/assessment-objectives) (aka grader guide). In every case, it should come right after the MongoDB lesson.

If you're not sure where to find these files, ask a Techtonica admin.

vegetabill:
Done!

PR #1170, opened 2020-01-26
Add mongoose outline to readme

alodahl:
@vegetabill are we making mongo an elective now? Based on how you re-arranged the readme, I did not add this lesson to the roadmap or the objectives.

addresses #967


vegetabill:
@alodahl thanks for fixing these. Yes Mongo is an elective now. I'd like them to learn SQL in more depth rather than NoSQL and SQL half as much. In terms of interviewing and career, SQL is still the dominant tech and learning NoSQL after a strong SQL foundation is easier IMHO than the reverse.

We really need prettier! Thanks for fixing my items. I think those came from the web editor on GitHub.

PR #1182, opened 2020-02-09
[API] Added "Intro to HTTP"

vegetabill:
There was no basic HTTP lesson and we got feedback that our foundational skills were missing so I split off and expanded the HTTP portions of my HTTP request practice lesson.


alodahl:
Does this need to be added to the roadmap or the readme?

PR #1183, opened 2020-02-09
[REST-API] rewrite REST lesson to be more clear

vegetabill:
I wasn't really able to follow this lesson so I simplified it.

### DIFF
--- a/api/REST-API.md
+++ b/api/REST-API.md
@@ -16,10 +16,7 @@ Here are links to lessons that should be completed before this lesson:
 
 
 ### Motivation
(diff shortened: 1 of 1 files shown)


vegetabill:
Thanks for taking a look @alodahl . The order is going to be:

1. Intro to HTTP (NEW)
1. HTTP Requests Practice
1. Intro to JSON (NEW)
1. APIs & JSON
1. AJAX (Skipped) - as HTTP requests covers it without unnecessary historical context

I'm going to update the roadmap and root README to match but if you have any feedback on the reordering, lmk here or in Slack.

PR #1186, opened 2020-02-11
React hooks lesson plan

dpenny:
**Questions**
How much React practice will they have between the Intro to React lesson and this?
Is this lesson an OK level of difficulty/detail?

### DIFF
--- a/react-js/react-hooks.md
+++ b/react-js/react-hooks.md
@@ -20,8 +20,8 @@ React is one of the most popular frameworks for front-end development. It is imp
 **Participants will be able to:**
 
 - Understand what pure functions and pure components are
-- Understand and implement the `useState` hook
-- Understand and implement the `useEffect` hook
+- Understand and use the `useState` hook
+- Understand and use the `useEffect` hook
 - Understand the benefits of using hooks
 
 ### Specific Things to Learn
@@ -74,7 +74,7 @@ Read [Intro to react hooks](https://reactjs.org/docs/hooks-intro.html) (5-10 min
 
 #### How do I use hooks?
 1. Read [Hooks Overview](https://reactjs.org/docs/hooks-overview.html)
-2. Walk through the examples on [Using the State Hook](https://reactjs.org/docs/hooks-state.html) to understand the state hook and its Class equivalent (in the counter `Example` component ).  Use [this codepen]([https://codepen.io/dpenny/pen/RwPPVqY](https://codepen.io/dpenny/pen/RwPPVqY)), to play around with these two ways of creating the same functionality. 
+2. Walk through the examples on [Using the State Hook](https://reactjs.org/docs/hooks-state.html) to understand the state hook and its Class equivalent (in the counter `Example` component ).  Use [this code sandbox](https://codesandbox.io/s/techtonicahooks0-8uy6j), to play around with these two ways of creating the same functionality. 
 
 *Discuss with a partner:*
 How do hooks achieve the same logic as the class component? Can you see how the code with hooks is cleaner?
@@ -84,41 +84,44 @@ How do hooks achieve the same logic as the class component? Can you see how the 
 
 A user should be able to:
 -  Add items (via text input field)
--  Press a button 'All done' which will clear all todo items, and a window alert will say 'Good job!'
+-  Press a button 'All done' which will clear all todo items, and a message on the top will say 'Good job!'
 - Press a button 'Clear all' which will clear all todo items
 - When there are more than 5 items on the list, there should be a status message on the top that says 'Try to complete current tasks before adding new ones! '
 This will walk through one possible approach to solving this using React hooks. Note that there is no one right answer or one right approach! 
 
-Please fork [this template](https://codepen.io/dpenny/pen/rNVVZrb) and follow along with this tutorial.
+Please fork [this template](https://codesandbox.io/s/techtonicatodotemplate-conbd) and follow along with this tutorial.
 
 0. **What should the component look like?**
 We can start with a 'sketch' of what our component might look like. This does not need UI or styling. We start with an outline so that we can think about what information and functionality this component will need. 
 
 For example,
-``` 
-const {useState, useEffect} = React;
-
+```
 const TodoApp = () => {
   return (
     <div>
-      <h3>status here</h3>
-      <div>
-        <input/>
-        <button>
-          Add item
-        </button>
-      </div>
-      
+      <b>status here</b>
+      <form>
+        <label>
+          New item:
+          <input type="text" name="name" />
+        </label>
+        <input type="submit" value="Add" />
+
+        <div>
+          <button> Clear all </button>
+          <button> All done </button>
+        </div>
+      </form>
+
       <h2> My Todos:</h2>
+      <ul>
         <li>item1</li>
         <li>item2</li>
-      
-
-      <button> Clear all </button>
-      <button> All done </button>
+      </ul>
     </div>
   );
 };
+
 ```
 
 2. **What information do we need? (aka what states do we need?)**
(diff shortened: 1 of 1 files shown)


alodahl:
Hi Danielle, this would come immediately after Intro to React. I'll take a look at this today or tomorrow to answer your second question.

react-js/react-hooks.md
```diff
@@ -2,89 +2,284 @@
...
+**Goal:  understand how we would use React hooks to implement a simple to-do list.**
+
+A user should be able to:
+-  Add items (via text input field)
+-  Press a button 'All done' which will clear all todo items, and a window alert will say 'Good job!'
+- Press a button 'Clear all' which will clear all todo items
+- When there are more than 5 items on the list, there should be a status message on the top that says 'Try to complete current tasks before adding new ones! '
+This will walk through one possible approach to solving this using React hooks. Note that there is no one right answer or one right approach! 
+
+Please fork [this template](https://codepen.io/dpenny/pen/rNVVZrb) and follow along with this tutorial.
```
gsong: This URL 404s.

gsong:
@dpenny @alodahl How do you feel about using https://codesandbox.io instead of CodePen for sandboxes?

1. It mirrors a real project a lot more.
2. It's VSC based, which most of the apprentices use, based on experience.
3. It prettiers code automatically (something they should get used to).

react-js/react-hooks.md
```diff
@@ -2,89 +2,284 @@
...
+#### Why were hooks introduced?
+Read [Intro to react hooks](https://reactjs.org/docs/hooks-intro.html) (5-10 minutes). 
+
+ Note:  React hooks were introduced in version 16.8. Previous alternatives to hooks include [render props](https://reactjs.org/docs/render-props.html),
+ [higher-order components](https://reactjs.org/docs/higher-order-components.html) such as the HOC used to connect a component to [the redux store](https://code.likeagirl.io/how-to-create-a-connected-hoc-with-react-and-redux-8631f7b1497). 
+ Right now you don't have to understand all of these alternatives in depth. The main point to note is there were a lot of solutions for the same problem.Hooks introduce *one* concept that solves these problems. 
+
+#### How do I use hooks?
+1. Read [Hooks Overview](https://reactjs.org/docs/hooks-overview.html)
+2. Walk through the examples on [Using the State Hook](https://reactjs.org/docs/hooks-state.html) to understand the state hook and its Class equivalent (in the counter `Example` component ).  Use [this codepen]([https://codepen.io/dpenny/pen/RwPPVqY](https://codepen.io/dpenny/pen/RwPPVqY)), to play around with these two ways of creating the same functionality.
```
gsong: @dpenny What's the best way to provide feedback on the code samples?

gsong:
@dpenny I opened a dpenny/curriculum#1 for formatting feedback, as they are too difficult to express here.

react-js/react-hooks.md
```diff
@@ -2,89 +2,284 @@
...
-
-Here are links to lessons that should be completed before this lesson:
-
-- Example: [Git version control](../git-version-control/git-version-control.md)
+ [React JS Part 1](https://github.com/Techtonica/curriculum/blob/master/react-js/react.md)
 
 ### Motivation
 
-Here's why this topic should be learned. Strengthen this section by including a company that uses this skillset, and a screenshot example if possible.
+React is one of the most popular frameworks for front-end development. It is important to have a way to store and change component state that is efficient and easy to read. Components can have their own state, or use shared stateful logic. Before hooks, there were many different approaches to sharing stateful logic. Hooks are great because they introduce *one* common way to reuse code. They ["let you split one component into smaller functions based on what pieces are related".]([https://reactjs.org/docs/hooks-intro.html#complex-components-become-hard-to-understand](https://reactjs.org/docs/hooks-intro.html#complex-components-become-hard-to-understand))
```
gsong: The quote "let you split..." felt abrupt to me in this context. I wasn't sure what it means. Perhaps just link to the relevant section in the docs and say "for more info, see..."?

react-js/react-hooks.md
```diff
@@ -2,89 +2,284 @@
...
 
 ### Objectives
 
 **Participants will be able to:**
 
-- leave this lesson knowing this
-- and how to do this on their own
-- Even more things
-- Even more things
+- Understand what pure functions and pure components are
```
gsong: What do you feel is the relevance of pure functions and pure components are, with regards to hooks?

react-js/react-hooks.md
```diff
@@ -2,89 +2,284 @@
...
+- Understand and implement the `useState` hook
```
gsong: We're not implementing the hooks, right? We're simply using them?

react-js/react-hooks.md
```diff
@@ -2,89 +2,284 @@
...
-	- This is a sub-thing
-- More things
-- Even more things
-- Even more things
+### Guided Practice
+**Goal:  understand how we would use React hooks to implement a simple to-do list.**
+
+A user should be able to:
+-  Add items (via text input field)
+-  Press a button 'All done' which will clear all todo items, and a window alert will say 'Good job!'
```
gsong: This isn't a valid requirement anymore, right? The status is displayed on the page now?

react-js/react-hooks.md
```diff
@@ -2,89 +2,284 @@
...
+This will walk through one possible approach to solving this using React hooks. Note that there is no one right answer or one right approach! 
+
+Please fork [this template](https://codepen.io/dpenny/pen/rNVVZrb) and follow along with this tutorial.
+
+0. **What should the component look like?**
+We can start with a 'sketch' of what our component might look like. This does not need UI or styling. We start with an outline so that we can think about what information and functionality this component will need. 
+
+For example,
+``` 
+const {useState, useEffect} = React;
```
gsong: This is a case where CodeSandbox is more like "real life." We would have an import statement instead. This isn't normal convention in typical React code or seen in official examples.

react-js/react-hooks.md
```diff
@@ -2,89 +2,284 @@
...
+
+const TodoApp = () => {
+  return (
+    <div>
+      <h3>status here</h3>
+      <div>
```
gsong: Do you think it would be better to model a proper `<form>`? I think semantic markup is super important, and we should model best practices.

react-js/react-hooks.md
```diff
@@ -2,89 +2,284 @@
...
+        <input/>
+        <button>
+          Add item
+        </button>
+      </div>
+      
+      <h2> My Todos:</h2>
+        <li>item1</li>
```
gsong: Hmm, `<li>`s without a list container. What do you think about adding `<ul>` container?

react-js/react-hooks.md
```diff
@@ -2,89 +2,284 @@
...
+          onChange={e => {
+	        // We want to set newTodo to be the input field value
+            setNewTodo(e.target.value);
+          }}
+        />
+        <button
+          onClick={() => {
+            // We want newTodo to be appended to items, and we want to setItems to be that new list. 
+            // We also want to reset the input field value
+            setItems([...items, newTodo]);
```
gsong: Hmm, this is the exact case that `useReducer` solves, I get we don't want to introduce that in the lesson, but I'm not sure modeling this way of using `useState` is the right thing to do either. Thoughts, @alodahl?

gsong:
@dpenny @alodahl I didn't have time to check on the quality of the linked materials, so @dpenny or someone else will have to vouch for them.

react-js/react-hooks.md
alodahl: Is it private, @dpenny?

vegetabill:
Hi @dpenny I realize there was a lot of review requests on this, being a brand-new lesson. Would it be possible to submit a revision by this weekend (the 29th)? The topic is fast-approaching on the cohort's calendar. 🤞

react-js/react-hooks.md
dpenny: in the original template the topic was 'hooks + pure functions'  @alodahl would you rather I move this section to react part 1?

react-js/react-hooks.md
dpenny: updated

react-js/react-hooks.md
dpenny: i wanted to start with a `useState` example because I think seeing different examples with `useState` will help them understand what it does (Even if useReducer is more efficient)- at the end there's a useReducer example, but I want them to feel comfortable with useState and useEffect before useReducer - thoughts?

react-js/react-hooks.md
vegetabill: I would just revise to be functional component vs class component. We don't need to cover pure here.

react-js/react-hooks.md
vegetabill: @dpenny this is still 404 so please make it shared

vegetabill:
I am merging this in as a draft since we need to start using it. I only skimmed @gsong comments because TL;DR. Feel free to turn any smaller items into GH issue(s). Especially with GSSoC starting now, we'll have a lot of capacity for simpler fixes.

PR #1245, opened 2020-03-04
Add Employees exercise for Recursion chapter

meherkasam:
Some folks expressed that they could use more practice with recursion. I prepared this diagram that you can use to build a tree of objects and do interesting things, both recursive and non-recursive on them. What I like about this is that it’s easy to visualise in the real-world. I’m hoping that the recursion will come out organically as you try to tackle this problem. I’m also happy to help add more complexity into the problem over time if folks are interested.
image


Kundan28:
Could you please link the issue here in the PR ?

alodahl:
Hi @meherkasam , could you please link the issue you are solving or otherwise describe why you are making this change?  Otherwise we are not sure how to review. Also, are you part of gssoc20?

Kundan28:
@alodahl Have you assigned @meherkasam any issue ?
As I don't see any.

alodahl:
Hi @meherkasam , please give us some more context on this change in the next few days, or we will close the PR.  Also, the image you added is incorrectly linked.

PR #1270, opened 2020-03-14
Improve mocha-testing lesson (#1241)

### LINKED ISSUE #1241: Improve mocha-testing lesson
https://github.com/Techtonica/curriculum/blob/master/testing-and-tdd/mocha-testing.md

Under **Mocha vs. Jasmine** section, there is a sentence saying "So, why might you choose one **other** the other?". The highlighted word "other" in this sentence could be replaced by "over" which sounds more appropriate.

 I am a GSSoC 20 participant and I would like to work in this issue.

## Additional Changes
- Add a note in the "should" section of Chai that you if you like that style, you can use the  [should library](https://www.npmjs.com/package/should) with mocha, as that is a common reason to use it over Jasmine
- Move this lesson to `/electives` folder
  - update its link in the root README after moving
- Include an async example in Chai and Mocha and remove the incorrect [Misconception No. 2](https://github.com/Techtonica/curriculum/blob/master/testing-and-tdd/mocha-testing.md#common-mistakes--misconceptions)
- Add a challenge: do the Mocha or Chai projects have their own internal tests? Suggest the student read them and see how they are used

Tahanima:
Closes #1241 

- Fixed typo
- Added async example for Mocha
- Added note under `should` section
- Improved indentation
- Removed Misconception no. 2
- Added challenge
- Moved the lesson to `/electives`
- Updated internal links
- Updated `README.md`

Please review the changes and let me know if there is anything I need to add/change.


electives/mocha-testing.md
```diff
@@ -122,6 +122,28 @@ Mocha String Test
     at Context.<anonymous> (test/myTest.js:5:16)
 ```
 - When calling `assert.equal(a, b)` mocha expects the values to be equivalent, or else the test will fail.
+- If you need to test asynchronous code, Mocha makes it very simple using callbacks. All you need to do is pass a callback function (usually named `done`) to `it` function and invoke the callback function after your test is complete. 
+```javascript
+var assert = require('assert');
```
alodahl: Isnt  assert from mocha or chai? Or is the package really called ‘assert’?

electives/mocha-testing.md
```diff
@@ -179,39 +201,37 @@ foo.should.equal('bar');
    - Jasmine is an out-of-the-box library that has test structure, assertions, displays test results, and spies.
    - Mocha is an extensible testing framework that provides a lot of flexibility
    - Chai is an assertion library that can be used with Mocha.
-2. Assertions inside of asynchronous code is ignored, therefore passing. This can lead to a false positive.
```
alodahl: This isn’t totally true, but can we replace it with a way to test async code?

electives/mocha-testing.md
```diff
@@ -122,6 +122,28 @@ Mocha String Test
     at Context.<anonymous> (test/myTest.js:5:16)
 ```
 - When calling `assert.equal(a, b)` mocha expects the values to be equivalent, or else the test will fail.
+- If you need to test asynchronous code, Mocha makes it very simple using callbacks. All you need to do is pass a callback function (usually named `done`) to `it` function and invoke the callback function after your test is complete. 
+```javascript
+var assert = require('assert');
+
+function asyncFunction(stringToTest, callback) {
+   setTimeout(function() {
```
alodahl: Can you warn them that this is not best practice and why?

electives/mocha-testing.md
```diff
@@ -240,7 +260,7 @@ describe("fizzBuzz", function(){
 
 ### Challenge
 
-No challenge yet!
+Do the Mocha and Chai projects have their own internal tests? Read them and see how they are used.
```
alodahl: Is part of the challenge to find the docs or codebase too? If not, please give them some links for where to start.

Tahanima:
Hello @alodahl 
I made the requested changes. Please review the changes and let me know if there is anything I need to add/change.

electives/mocha-testing.md
Tahanima: Done

electives/mocha-testing.md
Tahanima: Done

Kundan28:
Updated in scoresheet.

Issue #1359, opened 2020-05-15
Revise React Part 2 lesson


alodahl:
This issue requires basic markdown and React knowledge.

https://github.com/techtonica/curriculum/blob/master/react-js/react-part-2.md

- [x] on line 81, fix the example from using this.state to this.setState.
- [ ] Correct any grammar mistakes
- [x] Correct any md mistakes
- [x] Correct any code mistakes

nandanasreeraj123:
Can i do it?

Kundan28:
@nandanasreeraj123  If this is your first contribution then please go through [this](https://bit.ly/2wQxqLy) and fill out the form. And after that you could start contributing to this issue.

PR #1360, opened 2020-05-15
Line 81 rectified and grammatical mistakes cleared

nandanasreeraj123:
This PR rectifies issue #1359 
Grammatical errors were also modified.

- [x]  on line 81, fix the example from using this.state to this.setState.
- [x]  Correct any grammar mistakes
- [x]  Correct any md mistakes
- [x]  Correct any code mistakes

### DIFF
--- a/react-js/react-part-2.md
+++ b/react-js/react-part-2.md
@@ -14,7 +14,7 @@
 
 ### Motivation
 
-In the introduction to [React](/react.md), we discussed how React makes websites interactive and removes the need for jQuery code.  
+In the introduction to [React](/react.md), we have discussed how React makes websites interactive and removes the need for jQuery code.  
 Now the problem arises: How do we store and change component information we want in the UI? For example, maybe you want a button's background color to change once it's been clicked. The solution is to use React `state`.
 
 ### Objectives
@@ -30,7 +30,7 @@ Now the problem arises: How do we store and change component information we want
 
 ### Common Mistakes & Misconceptions
 
-1. The scope of the state is local to the component. This means that the state is restricted within class only.
+1. The scope of the state is local to the component. This means that the state is restricted within the class only.
 - Incorrect 
 	```javascript
 	class Message extends React.Component {
@@ -70,7 +70,7 @@ Now the problem arises: How do we store and change component information we want
   });
   ```
 
-3. A component should not change update its props. Props are values send by parent component to child component and changing props will cause inconsistency in prop data. Therefore it is advisable, not to change its own props.
+3. A component should not change or update its props. Props are values sent by parent component to child component and changing the props will cause inconsistency in prop data. Therefore, it is advisable not to change its own props.
 
 - Incorrect
   ```javascript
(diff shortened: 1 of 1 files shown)


react-js/react-part-2.md
```diff
@@ -108,12 +110,12 @@ Now the problem arises: How do we store and change component information we want
       message: 'new message'
     });
     ```
-- The important thing to remember about `setState` is that is causes the component to re-render. Well, in introduction to React, we have learnt that React follows a special path called the **React cycle**. So, when state is updated using `setState`, React goes into the **React cycle** and then the changes are compared from the previous **React virtual DOM element**, changes are done accordingly and those changes are reflected in the UI.
+- The important thing to remember about `setState` is that it causes the component to re-render. Well, in introduction to React, we have learnt that React follows a special path called the **React cycle**. So, when state is updated using `setState`, React goes into the **React cycle** and then the changes are compared to the previous **React virtual DOM element**. The changes are made accordingly and are reflected in the UI.
```
Kundan28, line 113:
```suggestion
- The important thing to remember about `setState` is that it causes the component to re-render. Well, in introduction to React, we have learnt that React follows a special path called the **React cycle**. So, when state is updated using `setState`, React goes into the **React cycle** and then the changes are compared from the previous **React virtual DOM element** and the changes are made accordingly and are reflected in the UI.
```

react-js/react-part-2.md
```diff
@@ -130,13 +132,13 @@ constructor(){
...
 
-3. A simple tag to display the number i.e. value defined in state.
+3. Create a simple tag to display the number i.e. the value defined in state.
 
 ```html
 <p>{this.state.value}</p>
 ```
 
-4. That is all which will be displayed on the page. Now we need to implement the functionality when the buttons are clicked.
+4. That is all what will be displayed on the page. Now we need to implement the functionality when the buttons are clicked.
```
Kundan28, line 141:
```suggestion
4. That is all which will be displayed on the page. Now we need to implement the functionality when the buttons are clicked.
```

react-js/react-part-2.md
```diff
@@ -308,4 +310,4 @@ UI should contain -
 
 - [How setState works](https://css-tricks.com/understanding-react-setstate/) - In depth knowledge of reconcilation process of React.
 - [Update state](https://medium.com/@baphemot/understanding-reactjs-setstate-a4640451865b) - State can be updated in two ways, either by passing an object or by passing function as a parameter.
-- [State in functional component](https://reactjs.org/docs/hooks-state.html) - So far we have seen how to use state in class components. State can be defined in functional components as well by using additional package called Hooks.
+- [State in functional component](https://reactjs.org/docs/hooks-state.html) - So far, We have seen how to use state in class components. State can be defined in functional components as well by using additional package called Hooks.
```
Kundan28, line 313:
```suggestion
- [State in functional component](https://reactjs.org/docs/hooks-state.html) - So far, we have seen how to use state in class components. State can be defined in functional components as well by using additional package called Hooks.
```

nandanasreeraj123:
changes have been made!

Kundan28:
Updated in scoresheet too.

nandanasreeraj123:
Thanku

PR #1370, opened 2020-05-21
Added js-dom-methods.md Fixes #1367

### LINKED ISSUE #1367: Create a vanilla JS lesson modeled after our jQuery one
Task: Create a JavaScript - HTML DOM Methods lesson modeled after our jQuery one.

Do not change the jQuery file; rather:
-  create a new file in web/ called js-dom-methods.md, 
- copy the contents of this [intro to jQuery](https://github.com/Techtonica/curriculum/blob/master/web/jquery.md) outline, 
- and then replace the content with JS DOM methods info instead.  
- Of course, if there are basics you think they should know about JS DOM methods please add them, and if there is nothing that corresponds to a detail about jQuery in vanilla JS, you can just remove that part.

Related to epic #1296

Mighty-Geek:


Mighty-Geek:
Thank you @vegetabill

Issue #1358, opened 2020-05-10
Follow a Heroku tutorial yourself and note problems you encounter


alodahl:
Task:  follow a Heroku tutorial yourself, actually follow the directions, open all the links, and write down any problems you encounter.

Work through this tutorial about deploying with Heroku: https://github.com/Techtonica/curriculum/blob/master/deploying/deploying-to-heroku.md

We need to find out what the problems are with it, and whether any of the directions or materials are outdated.  Whatever problems you find, please add them in the comments below.

Tahanima:
Hello @alodahl 
I am a GSSoC 20 participant and I would like to work on this issue.

Codess08:
Can I work on this issue?

alodahl:
@Tahanima make sure to follow these directions and then you can start:https://github.com/Techtonica/curriculum/blob/master/CONTRIBUTING.md

Tahanima:
@alodahl I got extremely busy and could not find time to work on this issue. Please unassign me. I apologize for any inconvenience caused.

alodahl:
@Codess08 are you still interested?

Codess08:
I'm afraid but this link gives "Page not found".
https://github.com/Techtonica/curriculum/blob/master/deploying/deploying.md

alodahl:
Check the deploying folder - we may have just changed the name to deploying-to-heroku or something similar.

alodahl:
https://github.com/Techtonica/curriculum/blob/master/deploying/deploying-to-heroku.md

alodahl:
Hi @Codess08, submit your review and your placeholder PR as soon as possible so we can review it and still have time for changes before the deadline.

thisLinda:
I would like to work on this project, it sounds pretty exciting. As an advanced beginner I've participated in MANY tutorials and so often wish I could give feedback on the process. There are times when the instructions gloss over a step or miss a small step because the author is familiar with what is being explained. I notice and admire when it's done well.

alodahl:
Yes, this is one of the advantages of having a junior on your team 😄
Thanks for offering to help!

thisLinda:
**Prerequisite** links are all active.

**Materials** links all active. Notes below regarding content of individual resources in Materials.

Video "What is Heroku," no issues.

Techtonica Slides: Deploying
Slide 6, FYI, refers to only PAAS and IAAS, video on slide 7 refers to SAAS, PAAS, and IAAS.
[...]

PR #1402, opened 2020-10-07
Issue #1358

thisLinda:
Minor changes (formatting, punctuation, instruction to student). UPDATE: I would like to revert this pull request but I do not see that it is possible.


thisLinda:
close request secondary to error

Issue #1204, opened 2020-02-26
Remove SQLite from curriculum and replace with Postgres


vegetabill:
Originally, the curriculum included some activities that used SQLite. This only saved a small amount of installation time and just caused confusion. Since SQL is further along in the curriculum, the participants should be capable of setting it up so let's go all in on Postgres.

## Requirements
- Find all references to SQLite
- Change them to use Postgres tooling and syntax

We will continue to rely on the [existing install instructions](https://github.com/Techtonica/curriculum/blob/master/databases/installing-postgresql.md). You can ignore the docker debate below.

For FAQs on how to contribute, see: https://github.com/Techtonica/curriculum/blob/main/CONTRIBUTING.md

vegetabill:
@alodahl is this something that was discussed and discarded before? Since the original curriculum used Mongo exclusively, not sure if this was really thought about in-depth.

gsong:
@vegetabill As long as by setup you mean https://hub.docker.com/_/postgres 😉

vegetabill:
😐

gsong:
@vegetabill Not a fan of Docker? Any specific reasons why?

vegetabill:
@gsong I like Docker just fine but we're trying to limit the moving parts here. Installing psql takes about 5 minutes using Homebrew so not sure how we would justify using Docker there unless we plan to Docker-ify the whole laptop environment. If you think that's worthwhile, I'd open a Epic to do that for future discussion. Having the laptops all set up the same way would be nice.

gsong:
My personal opinion is that a Docker based environment is much more predictable than your typical developer's local environment.  PostgreSQL is definitely easy to install via Homebrew, but it introduces its own challenges compared to isolated Docker DB instances (e.g. you can completely isolate your test DB from your development DB, no need to worry about user names, roles, etc. for different projects).

I also think there's value in setting a development environment that's closer to how a modern development environment would be set up.

No right or wrong answer here, and you're closer to the ground. So ¯\_(ツ)_/¯.

alodahl:
@vegetabill could we add this to gssoc?

alodahl:
@vegetabill I'm having a hard time understanding your original question. I'm fine to go forward with this issue.

alodahl:
@gsong as we have a limited time, containers won't make the short list I'm afraid.

vegetabill:
[...]

PR #1411, opened 2020-11-10
Change SQLite references to PostgreSQL

GazKris:
Addresses https://github.com/Techtonica/curriculum/issues/1204 by changing SQLite references and tooling to PostgreSQL.
In Addition:
https://github.com/Techtonica/curriculum/commit/387cf75ff4a8891168d33a709ac96091a3906956 updates the columns for the database to match the CSV pulled from the website in the instructions for backend.md.
https://github.com/Techtonica/curriculum/commit/5dc370e748dbf93dff9aab96568265f63dd8e6f7 fixes a typo in the reference to a previous step and corrects the link to the prereq lesson.

### DIFF
--- a/databases/sql.md
+++ b/databases/sql.md
@@ -18,7 +18,7 @@ Most companies use relational databases to store their data.
 
 **Participants will be able to:**
 
-- Interact with SQLite through the command line and through a GUI SQLite browser such as [DB Browser for SQLite](https://sqlitebrowser.org/)
+- Interact with PostgreSQL through the command line
 - Create database tables
 - Add, update, and delete data
 - Query data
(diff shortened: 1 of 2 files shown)


falun (approved):
The sqlite -> postgres conversion looks accurate. I didn't validate the contents of the files vs the new column definitions in the tables.

Caveat: I did this as read through + doc checking, did not actually execute the commands is a psql environment.

alodahl:
If the goal is to address (but not resolve) #1204 , then I approve.

GazKris:
I haven't gone through all of the google doc links in the curriculum yet, but I still plan to do this. As far as the md documents, I only went through the ones being used in the curriculum outline, but I am happy to go through the rest of the files, also. Since the sql.md page is now Postgres instead of sqlite, it will need to be moved to after the Postgres installation lesson, which I am also happy to do. 

I tested the Postgres code in my Ubuntu VM and they all worked, if that helps.

PR #1413, opened 2020-11-18
fix: error in the Guided Practice section of the REST-API lesson

### LINKED ISSUE #1412: Error in the Guided Practice for the REST-API lesson
I notice there is an error in the `PART II: Guided Practice` section:

`Now for the good part! Let's code together and build a small RESTful API, which will store and change data in a database instead of using hard-coded data within the project files.`

should be something like:

`Now for the good part! Let's code together and build a small RESTful API, which will store and change data which is hard-coded within the project file. This will mimic changes that would normally occur in a database, should the API be used in a real application.`

since we are **not** dealing with a database.

sylvaindessureault:
This fixes #1412

### DIFF
--- a/api/REST-API.md
+++ b/api/REST-API.md
@@ -145,7 +145,7 @@ DELETE	/customers/{id} : Delete customer with "id"
 
 ## PART II: Guided Practice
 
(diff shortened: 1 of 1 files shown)


vegetabill (approved):
Thanks for fixing this!

Issue #1332, opened 2020-04-09
Fix these dead links


alodahl:
Task: read through these notes describing the repo’s broken links, and repair or replace them.

react-js/react-hooks.md
15:12-15:93 warning Link to unknown file: version-control/git-version-control/git-version-control.md. Did you mean ../git-version-control/git-version-control.md missing-file remark-validate-links
***NO SUCH LINK anywhere in repo
redux/redux.md
***these LINKS ARE ALL ACTIVE but they are in react-js/react.md file (redux/redux.md no longer exists)

Note: Some  of the security links were fixed in #1397 
 the Activision link is old.
roles-in-tech/security.md
3:1-3:53 warning Link to unknown file: \roles-in-tech\security-locks.jpg missing-file remark-validate-links
***link now in career/roles-in-tech/security.md BAD/INCOMPLETE LINK for .jpg
17:268-17:383 warning Link to https://www.thenetworkpro.net/2010/04/02/why-is-computer-security-important/ is dead no-dead-urls remark-lint
***no such link appears to have been replaced with another, similar link
65:30-65:139 warning Link to https://wanguolin.github.io/assets/cryptography_and_network_security.pdf is dead no-dead-urls remark-lint
***NO SUCH LINK anywhere in repo
140:7-140:78 warning Link to https://www.themuse.com/jobs/eventbrite/security-engineer is dead no-dead-urls remark-lint
***DEAD LINK, now in career/roles-in-tech/security.md
148:78-148:175 warning Link to https://www.cloudflare.com/learning/security/glossary/what-is-penetration-testing/ is dead no-dead-urls remark-lint
***NO SUCH LINK anywhere in repo
196:3-196:48 warning Link to https://twitter.com/Mikab0nita is dead no-dead-urls remark-lint
***DEAD LINK, now in career/roles-in-tech/security.md
199:3-199:58 warning Link to https://live.ncsc.gov.uk/information/about-ncsc is dead no-dead-urls remark-lint
***DEAD LINK, now in career/roles-in-tech/security.md

abhu-A-J:
@alodahl Ma'am May I?

shreyakapoor08:
can I work on it? @alodahl

annu12340:
Can I work on this issue

Kundan28:
@abhu-A-J You could start working on this issue.

abhu-A-J:
> @abhu-A-J You could start working on this issue.

Okay Sure! Thank you!

[...]

PR #1415, opened 2020-11-22
Fix dead and incorrect links

CoderCarrot:
To close the issue #1332.
Most links were already fixed. 
https://www.cloudflare.com/learning/security/glossary/what-is-penetration-testing/ is live, but will continue to fail validation due to blocking programmatic access.


vegetabill (approved):
Thanks for fixing these!

PR #1416, opened 2020-11-29
SQLite to PostgreSQL

CoderCarrot:
To address issue #1204.
References to Postico added.
There is one reference to SQLite in the [Minimal API Project](https://github.com/Techtonica/curriculum/blob/main/projects/minimal-api-project.md) because the Sequelize package starts with SQLite. The reference states to change to PostgreSQL, so I imagine it's needed.


databases/sql-1.md
```diff
@@ -54,15 +54,18 @@ Most software applications store data of some sort -- user data, content, etc. D
 #### Section 1: Creating tables and basic querying
 
 1. Work through the [Codecademy SQL Tutorial](https://www.codecademy.com/learn/learn-sql) **Section 1 (Manipulation)**:
-2. SQLite is a version of SQL that comes installed on mac! Try it out:
-   - In Terminal, type `sqlite3 food.sqlite3` to use SQLite. This command will open sqlite and save your database to a file named `food.sqlite3`.
+
+2. PostgreSQL is a popular, open-source version of SQL. Try it out:
+   - In Terminal, type `pg_dump foo > food.psql`. This commande will save your database to a file names `food.psql`.
```
vegetabill, line 59: The equivalent for postgres is to create a new database. IIRC there should already be one named after their username which is fine for this exercise so we can probably just dive in.

databases/sql-2.md
```diff
@@ -45,24 +45,26 @@ Here's a sample schema:
         - which has an integer field named: `user_id`
         - and a text field named: `content`
 
-2.  Enter the SQLite command line program using the terminal commad `sqlite3 twitter.sqlite3`. (This will save your database to a file called `twitter.sqlite3`). Once there, use SQL statements to create two tables according to the spec for the two tables in step 1.
+2. Save your database by using the `command pg_dump twitter > twitter.psql`. (This will save your database to a file called `twitter.psql`.)
```
vegetabill, line 48: As before, this should probably be a CREATE DATABASE command.

databases/sql-2.md
```diff
@@ -45,24 +45,26 @@ Here's a sample schema:
...
     - All the tweets by a given user
     - The 10 most recent tweets by any user
     - Use a join to get a user's info along with their tweets
     - Make up your own!
 
-6.  Now let's perform a database migration. We will add a new column to an existing table in our database. Add a text field named: `bio` to your existing `users` table!
+7.  Now let's perform a database migration. We will add a new column to an existing table in our database. Add a text field named: `bio` to your existing `users` table!
 
-7.  Add some sample data to your new `bio` fields, but not for every user (leave some of their bios blank).
+8.  Add some sample data to your new `bio` fields, but not for every user (leave some of their bios blank).
```
vegetabill, line 65: [Optional] It's okay to leave them but for future reference, when making these numbered list items in markdown, it's actually better to make them all `1.` because that way they will be auto-numbered.

vegetabill (changes requested):
Yay this is great! Just a few small tweaks below and it'll be ready to merge.

CoderCarrot:
Thank you! I will be able to work on this this weekend, I believe.

databases/sql-2.md
CoderCarrot: If I understand correctly, you want them to use their DB named after their username that was created when installing Postgres for the table in these exercises? (The current commit reflects changes based on my understanding.)

vegetabill (approved):
Looks great! Thanks for revising.

PR #1438, opened 2021-01-04
Revise Eventonica Part 1 to match new requirements

### LINKED ISSUE #1420: [Eventonica] Revise Part 1 to match simplified reqs
Revise the [Part 1 Eventonica Lesson](https://github.com/Techtonica/curriculum/blob/main/projects/eventonica/eventonica-part1-objects.md) to match the simplified requirements in the [project README](https://github.com/Techtonica/curriculum/blob/main/projects/eventonica/README.md#project-requirements)

e.g. remove "search" and make the custom events favorites explicitly

vegetabill:
Fixes #1420 

Revise to meet new reqs and since we're trying to teach OO with this lesson, actually make a sensible OO skeleton.

### DIFF
--- a/projects/eventonica/eventonica-part1-objects.md
+++ b/projects/eventonica/eventonica-part1-objects.md
@@ -10,26 +10,15 @@ To build a modular system using Object Oriented Programming principles.
 
 ### Overview
 
-In this project, you'll build an event management system that will help you practice Object Oriented Programming using JavaScript.
+In this project, you'll build an event management system that will help you practice Object Oriented Programming using JavaScript. In the vocabularly of the popular [Model-View-Controller pattern](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller), this is the **Model**.
 
 ### How to get Started
 
 #### User Requirements
 
-You need to build a program that lets users browse a list of events and save the ones they are interested in. For today's piece of the project, you'll write object-oriented JS code to manage users and events. In a later lesson, you'll add a UI for your users to use.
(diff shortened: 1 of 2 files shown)


projects/eventonica/eventonica-part1-objects.md
```diff
@@ -39,52 +28,77 @@ Before you start coding, think about and write down the answers to these questio
...
   deleteUser() {
-    // Deletes a User from the system
+    // Deletes User
   }
+}
 
-  deleteEvent() {
-    // Deletes the Event from the system
+class Event {
+  static all = [];
```
alodahl, line 76: is `static` a JS thing?  I thought it was Java. I'm used to seeing `const` and `function`.

alodahl (commented):
one comment, otherwise this LGTM.

projects/eventonica/eventonica-part1-objects.md
vegetabill: Yep. Since this whole section is emphasizing OOP, it would actually probably be good to use it like other languages, since IIRC that was part of the point.

I'll add a link here:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/static

But maybe we should include that in one of the lessons? Not sure

alodahl (approved):
A link about `static` would be great!

projects/eventonica/eventonica-part1-objects.md
alodahl: I don't think we need it in the lessons - it's pretty obscure when it comes to JS development in my opinion.

PR #1442, opened 2021-01-19
Revise the partially complete React Part 3 - Components & Hierarchies…

### LINKED ISSUE #1336: React Part 3 - Components & Hierarchies
**Task**:  The previous contributor was unable to complete this issue, but they got quite far. Please:
- [ ] Make edits to this outline: https://github.com/Techtonica/curriculum/blob/master/react-js/react-part-3-component-hierarchies.md
- [ ] Make any changes needed so the Acceptance Criteria below is fulfilled in the outline.
- [ ] Address all the unresolved comments found here: https://github.com/Techtonica/curriculum/pull/1347

For FAQs on how to contribute, see: https://github.com/Techtonica/curriculum/blob/main/CONTRIBUTING.md
_____________________

## Original issue description:

_Child task of Epic #1301_

The previous two lessons focused on props and state inside of a single component. Now we need to expand on that foundation to cover how props and state are passed along component hierarchies and when it makes sense to split some JSX into a new component.

Use our [Lesson Template v2](https://github.com/Techtonica/curriculum/blob/master/_templates/lesson-v2.md)

## Topics to Cover
- Different between Components and built-in JSX tags like `<button>`
- How to refactor code where all JSX is in one component into separate components
  - Reasons to do so, duplication. The code exercise could be turning a map function that creates a snippet JSX to one that creates instances of another component
  - Keeping components small and single-purpose
- `props.children`
- Using 3p components like `reactstrap`
- Passing callback functions to have children allow the parent to decide what an onClick does
- Passing along state from a parent into a child component

## Example Code
- Example code should run in a CodePen or similar. It should not need Babel/Webpack, etc 
- Class components only (no Functional)

This lesson should go in the `./react-js/react-part-3-hierarchies.md` and a link added to the main README.md along with the other React lessons.

vegetabill:
… lesson

  - Address all open PR comments
  - Use women in computing quotes for test data
  - Clean up code into a new CodePen without using create-react-app or functional components
  - Closes #1336 (finally)

### DIFF
--- a/react-js/react-part-3-component-hierarchies.md
+++ b/react-js/react-part-3-component-hierarchies.md
@@ -21,7 +21,7 @@ React is a UI library. Building a React application involves breaking up your ap
 - Understand the difference between Components and built-in JSX tags
 - Understand how to pass data between components in react - using props and state
 - Understand the use of `props.children`
-- Understand how to use 3rd-party library of components like [reactstrap](https://reactstrap.github.io/)
+- Understand how to use a 3rd-party library of components like [reactstrap](https://reactstrap.github.io/)
 
 ### Materials
 
--- a/react-js/react-part-3-component-hierarchies.md
+++ b/react-js/react-part-3-component-hierarchies.md
@@ -12,7 +12,7 @@
 
 ### Motivation
 
-React is a User Interface (UI) library. Building a React application involves breaking up your app into smaller pieces called “components.” This make your code more readabel, manageable and scalable, as an app grows it can become difficult to get a sense of your overall component hierarchy and how components relate to each other. Sometimes, we want these components to communicate with each other and pass data among with each other. For this we use props and states. Sometimes, we would have to use 3rd party components to improve the the functionality or customization capability of out project
+React is a UI library. Building a React application involves breaking up your app into smaller pieces called “components.” This make your code more readable and maintainable. Usually we want these components to communicate with each other and pass data among with each other. For this we use props and state.
 
 ### Objectives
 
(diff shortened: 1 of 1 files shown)


react-js/react-part-3-component-hierarchies.md
```diff
@@ -12,16 +12,16 @@
...
 ### Objectives
 
 **Participants will be able to:**
 
 - Understand the difference between Components and built-in JSX tags
 - Understand how to pass data between components in react - using props and state
-- Understand what is props.children
-- Understand how to use 3rd party component like reactstrap
+- Understand the use of `props.children`
+- Understand how to use 3rd-party library of components like [reactstrap](https://reactstrap.github.io/)
```
alodahl, line 24:
```suggestion
- Understand how to use a 3rd-party library of components like [reactstrap](https://reactstrap.github.io/)
```

react-js/react-part-3-component-hierarchies.md
```diff
@@ -248,164 +229,153 @@ class App extends React.Component {
...
-1. QuoteText.js
-   This component takes in a text and wraps it in a h1 tags with an id "text" using an arrow function. It is then exported back
+#### Create the components for the quote generator app
 
-```js
-import React from 'react';
-const QuoteText = ({ text }) => <h1 id="text">{text}</h1>;
-export default QuoteText;
-```
+You will need to replace the placeholders above with actual `QuoteText` and `QuoteButton` components.
```
alodahl, line 279: Should there be a step reminding them to clone your starter pen and code along? Or are they just able to work in it directly (clearly I dont use codepen)

alodahl (commented):
Much better, thank you.

react-js/react-part-3-component-hierarchies.md
vegetabill: Haha I don't use it much either. Let me try it and see!

vegetabill: Looks like no. You can just start editing so I think we can just omit. Also by this point in the curriculum, they probably know more about how to use CodePen than me.

PR #1445, opened 2021-01-22
Update dot-profile.md

behnam:
Use one format for referring to `.zshrc` file.


alodahl (approved):
Much better, thank you!

PR #1446, opened 2021-01-25
Update dot-profile.md

behnam:
Use zsh-style PS1 parameters, instead of bash-style.


behnam:
Reference: https://scriptingosx.com/2019/07/moving-to-zsh-06-customizing-the-zsh-prompt/

alodahl (approved):

PR #1448, opened 2021-01-27
Update phase-3-bootstrap-prompt.md

aedward8:
As this is my first time using bootstrap I think it would be helpful to really walkthrough the steps and be explicit that <div class=”row”> needs to be added

### DIFF
--- a/projects/recipe-page/phase-3-bootstrap-prompt.md
+++ b/projects/recipe-page/phase-3-bootstrap-prompt.md
@@ -94,6 +94,8 @@ Save the file, then refresh `index.html` in Chrome by pressing `<COMMAND> + r`. 
 
 Do you remember how to tell Bootstrap to allow a div to take up a certain number of columns? On a small screen, we usually want divs to take up all 12 columns so things are as large as possible. Let's add the `col-xs-12` class to the divs that currently have the `top`, `left` and `right` classes. (Remember, elements may have more than one class!) Here's an abbreviated example:
 
+Don't forget to add your <div class=”row”> before your column <div>
+
 ```html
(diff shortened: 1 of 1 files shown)


alodahl:
@aedward8 could you please add a link to your description, and the exact section / line numbers that need to be changed?

alodahl:
@aedward8 are you requesting that we make an issue about this, or is all you want to change included in this PR?

projects/recipe-page/phase-3-bootstrap-prompt.md
```diff
@@ -94,6 +94,8 @@ Save the file, then refresh `index.html` in Chrome by pressing `<COMMAND> + r`.
 
 Do you remember how to tell Bootstrap to allow a div to take up a certain number of columns? On a small screen, we usually want divs to take up all 12 columns so things are as large as possible. Let's add the `col-xs-12` class to the divs that currently have the `top`, `left` and `right` classes. (Remember, elements may have more than one class!) Here's an abbreviated example:
 
+Don't forget to add your <div class=”row”> before your column <div>
```
alodahl, line 97: is this a note, or part of the code? Putting it after the "Here's an example:" is a little confusing -  Maybe add it to the other "remember" statement in the parenthesis just above?

alodahl (changes requested):

projects/recipe-page/phase-3-bootstrap-prompt.md
aedward8: Yes I think that suggestion will work! To just add it to the "remember statement"

alodahl (approved):

PR #1458, opened 2021-02-01
Update github-collaboration.md

aedward8:
I am a bit confused, for step 1 we forked and cloned our-user-name/Techtonica/curriculum to our local
by step 7 it wants us to also add Techtonica/curriculum remote to our folder. However because we already have the forked/cloned techtonica/curriculum remote it doesn't let us clone another repo with the same name "curriculum" I guess I am unsure why both repos which are basically the same need to be cloned. 
Unless I misunderstood the directions and we were not suppose to clone the fork repo from our own github


alodahl:
The point is not to clone both... let me think about a better way to describe it.

alodahl:
@aedward8 , does this help?: https://github.com/Techtonica/curriculum/pull/1473

alodahl:
I'm glad you brought this up. This seems a request rather than a proposed change -  a good thing to make an issue for.

alodahl:
@aedward8 would you say this is resolved?  If it is, you can make a comment , then close it with the button below.

aedward8:
Issue is resolved

PR #1462, opened 2021-02-03
Update personal-branding.md

lisaau:
Additional activity to add a profile README on participant's Github page as a way to build their personal brand.


vegetabill (approved):
Hey @lisaau this is great! I recently learned about that feature myself and this is a nice addition. Thanks for fixing the formatting as well!

PR #1465, opened 2021-02-10
Lesson slide video walkthrough is incorrect

aedward8:
The current Lesson slide video walkthrough is for a previous lesson and not this one.


vegetabill:
@aedward8 same with this one. Would you mind turning this into an issue?

PR #1470, opened 2021-02-13
Improve stalebot config

vegetabill:
- Use different rules for issues vs pull requests
  - Expand the stale window for issues to 180 days and exclude ones that have been "triaged" by the team and assigned to a milestone or project,  automatically without even needing "pinned" labels
  - For issues, add messaging encouraging them to make their own pull request
 
Config docs: https://github.com/probot/stale#usage


vegetabill:
@alodahl FYI I just merged this because the stale bot was still closing things. This should make it a lot more helpful and less spammy.

Issue #1296, opened 2020-03-19
[Epic] Eventonica v2021


vegetabill:
## TL;DR
High-level improvements to Eventonica based on H1-2020 cohort's experience along with Alina & Bill's Opinions.

## Background
In cohorts so far, the main project used to learn full-stack app development is [Eventonica](), a CRUD app for handling events of different categories. It was divided into [Parts 1-7](https://github.com/Techtonica/curriculum/blob/master/projects/eventonica/README.md).

## Eventonica Tenets
- Use the same reqs over and over to make it about the tech/technique, not the domain
- The more milestones the better, at least weekly
- Scheduling should occur such that Eventonica is reinforcing topics they have already covered at least once, although maybe not all combined in one task

## New Sequence

[Intro readme](https://github.com/Techtonica/curriculum/blob/main/projects/eventonica/README.md) should list user story-style features and these will be used throughout all the assignments.

| Topic                       	| H1-2020 Sequence 	| H1-2021 Sequence 🔼	| Issue(s)              	| Notes           	|
|-----------------------------	|:-----------------:	|:------------:	|-----------------------	|-----------------	|
| Models and Behavior         	|       Part 1      	|    Part 1    	|  #1420 ✅ 	|                 	|
| Jasmine Unit Testing                     	|       Part 2      	| Part 1.1 	|     -                  	|  	|
| Handwritten Web (HTML/CSS/JS DOM) 	|       Part 3      	|    Part 2    	|    #1419  ✅                 	|                 	|
| External Ticketmaster API   	|       Part 4      	|  - 	|  #1424                     	|  Feature cut               	|
| Express API (no DB)     	|       Part 5      	|    Part 3    	|   #1425                    	|                 	|
|  API Tests     	|       -      	|    Part 3.1    	|   #1215 WIP                    	|  Not a must-have. This is new and not essential               	|
| Backend uses Postgres                    	|       Part 6      	|    Part 4    	|   #1427                    	|                 	|
| React (client state only)           	|       Part 7      	|    Part 5    	|    #1421                   	|    Prereq: new React Part 4, #1337             	|
| Enzyme Testing           	|       -      	|    Part 5.1    	|   TBD - cut from 2021 scope                    	|     Practice writing Enzyme tests for one component      	|
| React calls API             	|         -         	|    Part 6    	|    #1428                 |  This was a really challenging and complex transition from React with client state only so we need some dedicated lessons/activities on this to ensure adequate prep. |

#### Optional Items
[...]

PR #1471, opened 2021-02-13
Update Eventonica/README to match the current state

vegetabill:
See epic #1296

We have been working on Eventonica changes directly on the main branch which has left things in a weird state for those trying to map out the schedule and link to resources.

This PR does all the renaming and reordering specified so that links will work even if the revisions aren't fully complete yet and notes which are not yet done for increased visibility.


naomiquinones (approved):
LGTM!

PR #1473, opened 2021-02-14
fix: clarify add remote directions

alodahl:
a fix for the issue mentioned here: https://github.com/Techtonica/curriculum/pull/1458


aedward8 (commented):
Yes,  the additions look good and are clear. Thank you for taking the time and making these adujustments

PR #1476, opened 2021-02-16
Update eventonica-part1-objects.md

aedward8:
I wanted to add a hint that I think is fundamental in understanding static methods. It took me a while to realize this and I think can point people in the right direction

### DIFF
--- a/projects/eventonica/eventonica-part1-objects.md
+++ b/projects/eventonica/eventonica-part1-objects.md
@@ -105,7 +105,7 @@ class User {
 
 ### What's with the `static` keyword?
 
-It's a way to include data that is shared across all instances of your class.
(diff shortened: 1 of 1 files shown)


alexcheva:
I found this on Codecademy: 
Within a JavaScript class, the static keyword defines a static method for a class. Static methods are not called on individual instances of the class, but are called on the class itself. Therefore, they tend to be general (utility) methods.
I think that might be a bit more clear.

naomiquinones (changes requested):
Let's add that logic as well as data is shared with the `static` keyword. Adding an example could help clarify this even further.

vegetabill:
Oof sorry all I just threw that static in without considering it enough.

Closing this for now. See #1477

PR #1477, opened 2021-02-17
Update "static" keyword definition.

alexcheva:


naomiquinones (changes requested):
@alexcheva, since this is a quote from Codecademy, you'll need to give proper attribution to Codecademy. Alternately, try rewriting it in your own words. You could also give an example of when a utility class method would be called.

vegetabill:
Adding the static was a mistake by me since I thought it was in the curriculum so we're actually cutting this entirely.

Thank you for being proactive and improving this in the meantime, @alexcheva

PR #1479, opened 2021-02-18
Update jasmine-testing.md

aedward8:
when I ran jasmine in CL when I am in the spec folder where my tests are,  it will say "Incomplete: No specs found". But If i run it in the folder just outside of spces jasmine-practice then it is able to see the files


vegetabill:
Thanks Abby!

PR #1482, opened 2021-02-18
Part of the solution is not hidden in details #9

aedward8:
I added an indent where the solution should still be hidden in the details since right now it is half exposed


vegetabill:
:shipit: I don't think I even knew Markdown did this

Issue #1215, opened 2020-02-28
Create API testing activity


vegetabill:
## Related Issues

- #1216 - We should use `supertest` since we already have some materials/lessons on it (and Bill thinks it's pretty good)
- #1296 - Eventonica Epic

## Description

Once Eventonica Part 5 is done, having API tests is a great way to provide stability while working on migrating to a database-backed backend. Let's make an in-class pair activity guide for it!

In the H1-2020 cohort, I just linked them to a short blog: https://codeforgeek.com/unit-testing-nodejs-application-using-mocha/

### Requirements
- Please add this project activity in the `projects/eventonica` folder.
  - See [project template](https://github.com/Techtonica/curriculum/tree/main/_templates) in our templates folder
  - Code examples should match the [Eventonica Domain](https://github.com/Techtonica/curriculum/blob/main/projects/eventonica/README.md) domain
- Include tests for GET, POST
- Include sample code for a single test case that creates data, verifies it exists, deletes it, and verifies it's gone
- For the Application Under Test (AUT) you can use [Lisa's code](https://github.com/lisaau/techtonica-assignments/tree/c194870fecadf0627def4b035a857c7dd7c3f846/Eventonica-with-Express) but note that the feature set has changed since she wrote it, so a few of the API endpoints wouldn't match. To minimize future rework, focus on endpoints that wouldn't change, e.g.
  - GET user
  - POST user

### Preferred
- Instead of using Mocha (an elective framework), preferably use node Jasmine to run these tests
  - I couldn't find any existing blogs for this so you might have to try it out on your own
- For the request-making, preferably use supertest (see related issue above)
- As a challenge, linking to how to leverage package.json's "scripts" feature to run all the tests or  start their server would be helpful

For FAQs on how to contribute, see: https://github.com/Techtonica/curriculum/blob/main/CONTRIBUTING.md

vegetabill:
@lisaau would you share your most complex test case code snippet here please?

lisaau:
Here's one I have to test POST: 

`describe("POST", function() {
    it("POST/user should add a user",function(done){
        let data = {userName: 'test', userID: 10};
        server
        .post('/user')
        .set('Content-type', "application/x-www-form-urlencoded")
        .send(data)
        .expect("Content-type",/text\/html/)
        .expect(200)
        .then( () => {
            server
            .get('/users')
            .end(function(err, res) {
                // check that user is added to the end of the array
[...]

PR #1485, opened 2021-02-20
Eventonica API testing exercise

lemon-yellow:
This addresses https://github.com/Techtonica/curriculum/issues/1215

- Added an API testing exercise using supertest and jasmine
- Edited prior testing exercise to include a script in `package.json` - this was part of the assignment but it made sense to include it with the prior exercise's Jasmine setup

### DIFF
--- a/projects/eventonica/README.md
+++ b/projects/eventonica/README.md
@@ -47,4 +47,5 @@ You'll work on this project over the next several weeks. The features will remai
 
 ## Optional Assignments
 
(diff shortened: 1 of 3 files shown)


vegetabill:
Yay this is great @lemon-yellow . I made some minor fixes and conflict resolution fixes on your branch. Hope you don't mind. Just wanted to get this out ASAP so this cohort can choose to use it if they want. 

cc @SarahCawley

PR #1487, opened 2021-02-22
change typo nvm to npm

aedward8:
i believe its a typo from nvm to npm


aedward8:
nvm is actually nvm saw in the next part of the lesson

PR #1488, opened 2021-02-22
install with Brew or instructions in read me

aedward8:
These instructions don't go with the instructions in the readme. I tried to do both and it kinda mess things up. Eventually just went with what was exactly in the the readme so maybe makes sense to se updates these instructions to that? Also I didn't have permissions to with brew to download nvm so that threw a lot of errors and issues for me.


vegetabill:
Hi @aedward8 , can you open this as a new issue for an improvement request? Thx!!

PR #1489, opened 2021-02-22
The lab exercises

aedward8:
The lab exercise has us running node javascript code in our command line. Since we used the command line we did not use or create an app.js file. In the exercise we did not create a basic web app using Node.


vegetabill:
@aedward8 I think this was probably meant to be an issue but I pushed another commit that removed the inaccurate "Check for Understanding" with a generic one.

Thanks for reporting this issue!

PR #1494, opened 2021-02-23
propose to change requirement to install telnet

picaq:
telnet is deprecated and unsupported. Installing it on my machine yielded errors as shown in this thread
https://discussions.apple.com/thread/251160017


vegetabill:
Thanks for raising this issue!

On my machine running Big Sur 11.2.1 Homebrew doesn't mention any problems and it works fine.

However, reading around on the internet and it seems like netcat `nc` allows this as well and is already built in to Mac so I changed it to use that instead.

PR #1495, opened 2021-02-24
Challenge section was a bit confusing

aedward8:
The Challenge was very confusing as the regular way to access user.agent is through the window object => window.navigator.user-agent but this window object only exists in the browser and not server side node.  I was able to find this code in a gist that was able to be used 

https://gist.github.com/ninjascribble/5119003


naomiquinones (approved):
As per discussion with @Sarahcawley, this seems good to go

PR #1498, opened 2021-02-25
Update "Intro to HTTP" prerequisite link

carbonsoda:
The "How the Internet Works" topic currently points to `requests-and-response.md`, but the file was since renamed to `how-the-internet-works.md` and moved to the `networking-computing` directory


vegetabill (approved):
Thanks for this fix @carbonsoda

PR #1472, opened 2021-02-13
[Eventonica] Revise Part 7 - React

vegetabill:
Resolves #1421 

  - Fill in Eventonica Part 3 lesson with step-by-step examples of porting a JS DOM app to React
  - This should be rebased against #1471 first
  - Sample code adapted from @lisaau https://github.com/lisaau/Eventonica/blob/master/src/jQuery-scripts.js

PR #1505, opened 2021-03-01
[Eventonica React] - merge existing content with additional porting instructions

vegetabill:
Originally opened as #1472 but I messed up that branch in Git 😬 

- include specific walkthrough of porting JS DOM code => React
- Link to other new content since originally created
- Keep all server / API hints

### DIFF
--- a/projects/eventonica/eventonica-part7-react.md
+++ b/projects/eventonica/eventonica-part7-react.md
@@ -7,19 +7,22 @@
 ### Primary Goals
 
 - Get experience writing your own React code
-- By the end of this project, you should have a full-stack app you wrote including the database, backend, and frontend, all working together.
+- Compare and contrast the code when using React and not using React
 
-### Overview/Instructions
+## Instructions
 
 In this project, you'll create a React frontend for your Eventonica API. We can base the HTML off of your Part 2 HTML, but you will need to turn your code into React components.
 
-#### Step 1: Requirements
+### Requirements
 
 You will be using the same Express API from your existing Eventonica project. As always you will want to support the features covered in the main [Eventonica README](./README.md).
+As always, we will be working from the same set of [Eventonica Requirements](./README.md) as all parts of this project.
 
-#### Step 2: Set up React App
+### Set up a new React App with API Support
 
-##### Note: when following these steps you will now have **TWO SERVERS** running:
+We'll use `create-react-app` to set up a basic app, first covered in [React Full App](../../react-js/react-part4-full-apps.md) lesson.
+
+_Note: when following these steps you will now have **TWO SERVERS** running, one for your Express API and one to serve the live-reloaded React app._
 
 1. The first one is your existing Express server that talks to the database and serves your API routes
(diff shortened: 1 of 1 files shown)


projects/eventonica/eventonica-part7-react.md
```diff
@@ -72,10 +70,145 @@ Here's an example of what `package.json` might look like now:
...
+
+#### JavaScript
+
+You should be able to move any JavaScript files into the `public` folder and use them as-is. At this point your app should work but not really using React in any way. Make sure the basics work before proceeding.
+
+### Create Your First Component
+
+Before this stage, ensure you have a commit in place with the working app.
+
+Here is example code that might be similar to the way you displayed a list of users in your app. Depending on how you built Part 2, your code might not be in a function like this but hopefully you can translate the concept to your own code.
```
vegetabill, line 89: Fixed typos from #1472

projects/eventonica/eventonica-part7-react.md
```diff
@@ -72,10 +70,145 @@ Here's an example of what `package.json` might look like now:
...
+        <input id="delete-user-action" type="submit" value="Delete User" />
+      </form>
+    );
+  }
+}
+```
+
+### Re-rendering the UI in Response to Changes
+
+Now we want the UI to re-render when we delete the user. To do this we will have to call `this.setState` with updated data to ensure that React will re-render. Use the previous lessons you went through to figure out how to do this.
```
vegetabill, line 209: Please review this new wording, @alodahl . The last wording was confusing and came from me having mixed results teaching immutable code patterns as the participants might see in a real codebase vs. modifying and objects in state and then just calling `this.setState` with the whole object again.

projects/eventonica/eventonica-part7-react.md
```diff
@@ -72,10 +70,145 @@ Here's an example of what `package.json` might look like now:
...
+    displayEvents(); // refresh the DOM
+  });
+```
+
+In React we would setup the event handler like this:
+
+```jsx
+// src/UserDeleteForm.js
+class UserDeleteForm extends React.Component {
+  state = { idToDelete: null };
```
alodahl, line 182: Thanks for doing this!  Again, I'm not sure this is the way to initiate state... I think you need a constructor and everything.

alodahl (changes requested):

projects/eventonica/eventonica-part7-react.md
alodahl, line 182:
```suggestion
  constructor(props) {
  super(props);
  this.state = { idToDelete: null };
}
```

projects/eventonica/eventonica-part7-react.md
alodahl: One thing that's really important for React that you don't have to care about with jQuery is that the user ID list should be in the state of the parent component, meaning that the onClick will need to be a callback prop, and you probably want the callback function to live in the parent to make it clear.  If you think they've learned enough to accomplish this, maybe you can just leave a note reminding them of these things, but let them write the code themselves?

PR #1506, opened 2021-03-02
added bodyParser to express set up

SarahCawley:


naomiquinones (commented):
According to this [Stack Overflow](https://stackoverflow.com/questions/47232187/express-json-vs-bodyparser-json/47232318#47232318),

> bodyParser was added back to Express in release 4.16.0, because people wanted it bundled with Express like before. That means you don't have to use bodyParser.json() anymore if you are on the latest release. You can use express.json() instead.

So it seems like Express versions 4.16.0 and above don't need to require it explicitly.

Perhaps a comment could be added before or after the require line to specify that it should be added if using Express versions 4.0 through the one right before 4.16.0?

vegetabill:
TIL, awesome @naomiquinones

PR #1507, opened 2021-03-04
command to set new database in PostGres

LeiaQuesada:


LeiaQuesada:
oh wait, but this doesn't automatically set this as a super user? and how do I add a semicolon before we merge?

lalpert:
Thanks for working on this! I've created a PR to fix the semicolon issue: https://github.com/Techtonica/curriculum/pull/1510
As for the other changes, I'll think about how you should be connecting to postgres and follow up with more comments!

vegetabill:
FWIW, best practice is to connect as your own user (by not specifying any args to `psql`), which by default connects to a db with the same name which should have been created during installation via homebrew. 

In terms of the future, I would suggest we push more use of Postico, as even in a professional setting I rarely see engineers using command line `psql`. What do you think @lalpert ?

PR #1508, opened 2021-03-04
Need semicolon after create database

aedward8:
we went over with George.

Also George said that this is bad practice
"Now you can connect with just `psql`" - so maybe we wanna to take this out?

Also George said he doesn't recommend we name the server after our username but to our project - Eventonica

### DIFF
--- a/databases/installing-postgresql.md
+++ b/databases/installing-postgresql.md
@@ -16,9 +16,8 @@
 - This creates a superuser named postgres
 - If you run `psql` with no args, it assumes you want to connect as your current user (`whoami` command output)
 - Run `psql postgres` to connect as the superuser
-- `create database <YOUR_USERNAME>`
+- `create database eventonica`;
 - `\q` to quit
-- Now you can connect with just `psql`
 
 ### Configuration
 


lalpert:
You're totally right that we need a semicolon in that line. I need to think more about the other changes since I'm not convinced you should be doing the rest of the upcoming postgres exercises in your `eventonica` database, and we'd need to be sure to change the other lessons to match (like https://github.com/Techtonica/curriculum/blob/main/databases/sql-1.md)

In the meantime I created a PR to fix just the semicolon issue https://github.com/Techtonica/curriculum/pull/1510

aedward8:
Hi Leah that sounds good. Also I realized the benefit of naming the database our username was that it easily connects to postico. So maybe that was the reason.

I will also break up my PR from this point on to small individual sizes instead of combining multiple changes at once in case this sort of thing happens again.

Thank you!

vegetabill:
Abby, your guess is correct: it's for easy connection. Merging the semicolon fix!

> Also George said that this is bad practice

@gsong would love more context about this "bad practice".

You will create many databases during Techtonica, this is just to test the first one so I'm leaving it as written.

vegetabill:
Oh oops someone already merged it. Nvm!

PR #1512, opened 2021-03-05
change " to ' in pizza

aedward8:


akivalencia:
to clarify "pizza" in the SQL does not run. We must use '' single quotes.

vegetabill:
Good catch! Thanks for opening this fix.

For future learning, double quotes are valid in SQL but they are used to escape unusual table/column names, never for strings/varchar values.

PR #1513, opened 2021-03-05
add id as primary key for independent practice

aedward8:
Since we have been learning the importance of including a primary key in our table I updated the independent practice to include a primary key column

### DIFF
--- a/databases/sql-1.md
+++ b/databases/sql-1.md
@@ -57,7 +57,7 @@ Most software applications store data of some sort -- user data, content, etc. D
 
 1. PostgreSQL is a popular, open-source version of SQL. Try it out:
    - In Terminal, type `psql` to use PostgreSQL.
-   - Try adding a table with the command `CREATE TABLE food (name TEXT, calories INTEGER);`
+   - Try adding a table with the command `CREATE TABLE food (id INTEGER,name TEXT, calories INTEGER);`
    - See that the table was created by typing `.tables` to see the list of all existing tables.
    - Try adding some data: `INSERT INTO food VALUES ("pizza", 500);`
    - See the data: `SELECT * from food;`
@@ -69,7 +69,7 @@ Most software applications store data of some sort -- user data, content, etc. D
 1. Try writing queries for the following using the `food` table:
    - Add a column to the table called `meal`
    - Update "pizza" to have its meal be "dinner"
-   - Insert 4 more rows into the food DB. Be sure to include values for all 3 columns -- `name`, `calories`, and `meal`
+   - Insert 4 more rows into the food DB. Be sure to include values for all 4 columns -- `id`, `name`, `calories`, and `meal`
    - Update "pizza" to have a different calories number
    - Select only the names of all the foods
    - Make up a query of your own using another command you've learned


databases/sql-1.md
```diff
@@ -57,7 +57,7 @@ Most software applications store data of some sort -- user data, content, etc. D
 
 1. PostgreSQL is a popular, open-source version of SQL. Try it out:
    - In Terminal, type `psql` to use PostgreSQL.
-   - Try adding a table with the command `CREATE TABLE food (name TEXT, calories INTEGER);`
+   - Try adding a table with the command `CREATE TABLE food (id INTEGER,name TEXT, calories INTEGER);`
```
lalpert, line 60: Could you include a space after the comma here to be consistent?

lalpert (commented):
Good idea! I commented on one small change to make. And there's a git conflict so make sure to resolve that.

databases/sql-1.md
aedward8: Hi @lalpert  thank you for your comment I added the comma and spacing as well as fixed the conflict which was that I added `\dt` to display table.

databases/sql-1.md
```diff
@@ -55,21 +55,21 @@ Most software applications store data of some sort -- user data, content, etc. D
...
 1. Work through the [Codecademy SQL Tutorial](https://www.codecademy.com/learn/learn-sql) **Section 1 (Manipulation)**:
 
-1. PostgreSQL is a popular, open-source version of SQL. Try it out:
+2. PostgreSQL is a popular, open-source version of SQL. Try it out:
    - In Terminal, type `psql` to use PostgreSQL.
-   - Try adding a table with the command `CREATE TABLE food (name TEXT, calories INTEGER);`
-   - See that the table was created by typing `\dt` to see the list of all existing tables.
-   - Try adding some data: `INSERT INTO food VALUES ('pizza', 500);`
+   - Try adding a table with the command `CREATE TABLE food (id INTEGER, name TEXT, calories INTEGER);`
+   - See that the table was created by typing `.\dt` to see the list of all existing tables.
```
lalpert, line 61: There shouldn't be a `.` before `\dt`

databases/sql-1.md
```diff
@@ -55,21 +55,21 @@ Most software applications store data of some sort -- user data, content, etc. D
...
+   - Try adding some data: `INSERT INTO food VALUES ("pizza", 500);`
```
lalpert, line 62: double-quotes aren't the same as single-quotes in SQL so this needs to be changed back to single quotes.

Also, now that there is an "id" column in the food table, this insert statement should include a value for id as well.

You should be able to double-check that your changes are correct by running the new commands in SQL and seeing that they work -- let me know if you hit any problems doing that!

databases/sql-1.md
```diff
@@ -55,21 +55,21 @@ Most software applications store data of some sort -- user data, content, etc. D
 
 1. Work through the [Codecademy SQL Tutorial](https://www.codecademy.com/learn/learn-sql) **Section 1 (Manipulation)**:
 
-1. PostgreSQL is a popular, open-source version of SQL. Try it out:
+2. PostgreSQL is a popular, open-source version of SQL. Try it out:
```
lalpert, line 58: It's fine to leave in the re-numbered list items, but just fyi, markdown automatically puts in the right list numbers for you when rendering the page. People often use all 1's in markdown lists to make it easier to add intermediate steps later without needing to renumber the list.

lalpert (commented):
Thanks for making those changes! I see some other changes in this PR too now so I added comments on those.

databases/sql-1.md
aedward8: @alodahl  I believe this issue is resolved and ready for merge

PR #1517, opened 2021-03-05
Add a comment how there is psuedo code

aedward8:
Many of us got confused on how these routes were deleting /posting/putting not realizing it was psuedo code.

### DIFF
--- a/api/REST-API.md
+++ b/api/REST-API.md
@@ -445,6 +445,7 @@ app.route('/invoices/:id').get((req, res) => {
 If you test your routes with Postman or by opening a browser and going to `http://localhost:3000/customers/1` and to `http://localhost:3000/invoices/6`, you should see these 2 records being returned.
 
 Now, in order to have a full-out API, we need more routes. So, we will now create the remaining routes, mainly the `POST`, `PUT` and `DELETE` ones.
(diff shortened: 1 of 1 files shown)


vegetabill:
Thanks @aedward8 for capturing this problem and offering a fix!

It's been awhile since I've looked at this lesson but I believe the code is not meant to be pseudo-code but rather real code. I copy and pasted the snippets into a single `server.js` file and it seemed to work.

Do you remember what the error you encountered was?

aedward8:
Yes! for example the "Delete" section of the guided practice , the code I believe runs but it doesn't actually delete anything it just does a filter function which I believe will run but no actually deleting occurs. 

similarly for put and post its like the general structure of the code but not exactly 100% updating or adding new content. We also confirmed this with the SEAM (sarah)

vegetabill:
Ahhhh gotcha. Sorry, I only tried running some of it to catch errors. Obviously we didn't review this submitted lesson carefully enough.

Given that **Guided Practice** is not meant to leave that much unspecified and at a minimum should have `//your code goes here` if something is missing, I converted this to an issue: #1525

I merged your warning as a sign post for readers in the meantime and once someone resolves that issue we can remove your note which should no longer be accurate.

Thanks!

PR #1518, opened 2021-03-05
Also added a line about psuedo code

aedward8:
for clearer instruction, similar to what was in REST API  guided practice lesson

### DIFF
--- a/projects/mailing-list-rest-api.md
+++ b/projects/mailing-list-rest-api.md
@@ -69,7 +69,8 @@ app.get('/lists', (req, res) => {
   "members": ["talea@techtonica.org", "michelle@techtonica.org"]
 }
 ```
-Please note that the logic for `POST`, `PUT` and `DELETE` is psuedo code and won't actually run, similar to this lesson's guided practice. The goal of this exercise is for you to understand how to set up routes so don't worry too much about the logic yet, we will get to that in future lessons.
+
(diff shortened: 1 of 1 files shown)


projects/mailing-list-rest-api.md
```diff
@@ -69,6 +69,7 @@ app.get('/lists', (req, res) => {
   "members": ["talea@techtonica.org", "michelle@techtonica.org"]
 }
 ```
+Please note that the logic for `POST`, `PUT` and `DELETE` is psuedo code and won't actually run, similar to this lesson's guided practice. The goal of this exercise is for you to understand how to set up routes so don't worry too much about the logic yet, we will get to that in future lessons.
```
vegetabill, line 72: I'm not sure I see what pseudo-code this is referring to. Can you clarify, @aedward8 ?

vegetabill (changes requested):

projects/mailing-list-rest-api.md
SarahCawley: The backend portion of the exercise is not fully set up so the code will error. Maybe rephrase as something like "while the front end code is correct the backend is not set up and will error. Do not worry if you can not make it work, please follow along with the code and make sure you understand what is going on."

vegetabill:
Thanks @aedward8 @SarahCawley. I combined both of the info into one note. If it's not quite accurate, please lmk!

PR #1519, opened 2021-03-05
Add a note in the instruction about calls being in AJAX

aedward8:


vegetabill (approved):
Thanks @aedward8 . I slightly tweaked this to include a link to what the next lesson was. Let me know if I got it wrong lol...I went by Monday.com

PR #1535, opened 2021-03-11
removed update function in child component

aedward8:
i think the update function was left over c/p from the parent component and doesn't belong there


vegetabill:
Yep you're right! Thanks for fixing.

PR #1541, opened 2021-03-15
Update react-part-3-component-hierarchies.md

m3ia:
Broken link


vegetabill:
Thanks for reporting. Unlike a broken link pointing to a blog that no longer exists, we should recreate this and probably fork all these pens to a Techtonica-owned account. This is on the personal account of a random volunteer.

I've opened an issue about it: #1549

Merging this in the meantime to reduce confusion.

PR #1545, opened 2021-03-15
Remove duplicate Todo component, + info in parens

picaq:
Duplicated Todo component in App.js after it has already been imported from Todo.js and will generate an error when attempting to declare it again. Removing it from App.js resolves the error. The Todo app may not render without refreshing the chrome tab that the default App.js was rendered in ( probably due to cache ).

### DIFF
--- a/react-js/react-part4-full-apps.md
+++ b/react-js/react-part4-full-apps.md
@@ -110,13 +110,6 @@ import React from 'react';
 import Todo from './Todo';
 import './App.css';
 
-class Todo extends React.Component {
-  render() {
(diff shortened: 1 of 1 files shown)


akivalencia:
^^^^ no need to declare the class Todo twice. Just do not add it to App.js

naomiquinones (commented):
This does seem like an oversight. @vegetabill, what do you think?

vegetabill (approved):
Oops yes @naomiquinones just an oversight. This lesson is brand new this year so no one has used it yet for real.

PR #1554, opened 2021-04-07
newer-Js-features

renushinde:


renushinde:
some newer JS features including default params, spread and rest operator and destructuring assignment.

javascript/newer-JS-features.md
```diff
@@ -0,0 +1,219 @@
...
+
+### Motivation
+
+ECMAScript 2015, or ES6 introduced many important and useful features that modernized the JavaScript language and helps us to speed the development process. 
+
+### Objectives
+
+**Participants will be able to:**
+
+- Take advantage of ES6's new defalut params.
```
alodahl, line 25:
```suggestion
- Take advantage of ES6's new default params.
```

javascript/newer-JS-features.md
```diff
@@ -0,0 +1,219 @@
...
+- Use spread operator with arrays, functions and objects.
+- Know the difference between spread and rest operator. 
+- Use of Destructured Assignment
```
alodahl, line 28:
```suggestion
- Use of Destructuring Assignment
```
I've only heard it phrased as "Destructuring Assignment": https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

javascript/newer-JS-features.md
```diff
@@ -0,0 +1,219 @@
...
+- [The Modern JavaScript Tutorial](https://javascript.info/)
+- [ES6 Tutorial](https://www.youtube.com/watch?v=WZQc7RUAg18)
+
+
+### Lesson
+
+[Modern JS Features slides](https://docs.google.com/presentation/d/138TkIVXTaRS3ciGYZ8A62q1QVaoA0ACcTmqVDjFv8-4/edit#slide=id.gced263eb5d_0_167)
+### Things to Remember
+
+- **No browser fully supports ES6**, to check browser support [Can I Use](https://caniuse.com/) is a great resource.
```
alodahl, line 42: 💙

javascript/newer-JS-features.md
```diff
@@ -0,0 +1,219 @@
...
+What if, we forget to pass value for b? We'll get NaN.
+To avoid getting NaN, we can have a default value set for b. 
+```javascript
+function multiply(a,b) {
+    b = typeof b !== 'undefined' ? b:2;
+    return a * b;
+}
+multiply(5) // 10 
+```
+Here, I've used JS ternary operator to check the condition for b. If b has an undefined value, then thedefault value of 2 is assigned to b. When we call function, we'll get 10.
```
alodahl, line 66:
```suggestion
Here, I've used JS ternary operator to check the condition for b. If b has an undefined value, then the default value of 2 is assigned to b. When we call the function with 5, we'll get 10. 
```

javascript/newer-JS-features.md
```diff
@@ -0,0 +1,219 @@
...
+}
+multiply(5) // 50.
+```
+In the example above, if no value is passed for b, i.e if b is undefined, it'll get a default value of 10. This is a much nicer and shorter syntax. 
+
+## Spread Operator
+The Spread (...) operator is a useful and newer syntax for adding items to arrays, combining arrays or objects, and spreading an array out into a function’s arguments.
+## Some scenarios where Spread is useful:
+- Copying an array
+- Conncatenating or combining arrays
```
alodahl, line 81:
```suggestion
- Concatenating or combining arrays
```

javascript/newer-JS-features.md
```diff
@@ -0,0 +1,219 @@
...
+- Merge two Objects.
+
+## Examples
+### Spread in Function Calls
+We can use Math.max to get the greatest number from a list.
+```javascript
+Math.max(2,3,4,5,6) // 6
+```
+Here, Math.max ia expecting seperate arguments or values to be passed on and it'll give us the maximum value. But, what if we use Math.max on array?
```
alodahl, line 90:
```suggestion
Here, Math.max is expecting separate arguments to be passed in and it'll give us the maximum value. But, what if we use Math.max on an array?
```

javascript/newer-JS-features.md
```diff
@@ -0,0 +1,219 @@
...
+### Lesson
+
+[Modern JS Features slides](https://docs.google.com/presentation/d/138TkIVXTaRS3ciGYZ8A62q1QVaoA0ACcTmqVDjFv8-4/edit#slide=id.gced263eb5d_0_167)
+### Things to Remember
+
+- **No browser fully supports ES6**, to check browser support [Can I Use](https://caniuse.com/) is a great resource. 
+
+
+## Default Params
+The default param is a way to set default values when function parameter has no or undefined values.
```
alodahl, line 46:
```suggestion
The default param is a way to set default values when the function parameter has undefined values. 
```
I don't want to confuse anyone; no defined values === undefined.

javascript/newer-JS-features.md
```diff
@@ -0,0 +1,219 @@
...
+Math.max(...[2,3,4]); // 4 (spread expands array into seperate arguments)
+```
+### Spread with Arrays
+Spread in array literals creates a new array using an existing array. It basically spreads the elements from one array into a new array.
+Let's look an example -
+Here, I've 2 arrays primaryColors and secondaryColors. allColors array contain all the elemnts from primaryColors and secondaryColors. ... spread the values into seperate arguments. allColors contains copy of primaryColors and secondarycolors but the original arrays remains unchanged! 
+```javascript
+const primaryColors = ['red', 'green', 'blue'];
+const secondaryColors = ['purple','pink', 'crimson'];
+const allColors = [...primaryColors, ...secondaryColors] // (6)["red", "green", "blue", "purple", "pink", "crimson"]
```
alodahl, line 106: 👏🏼

javascript/newer-JS-features.md
```diff
@@ -0,0 +1,219 @@
...
+    favColor:"pink",
+    favPet:"cat"
+}
+const personD = {
+    favFood:"pasta",
+    favPet:"dog"
+}
+const anotherInfo = {...personA, ...personB } // {favColor: "pink", favPet: "dog", favFood: "pasta"}
+```
+## Rest parameters...
```
alodahl, line 132:
```suggestion
## Rest parameters
```
I don't understand how the ellipses are useful

javascript/newer-JS-features.md
```diff
@@ -0,0 +1,219 @@
...
+Rest parameters allows a function to accept an indefinite numbers of arguments as an array. Rest looks like spread, but it’s NOT spread. Before taking a look at Rest, let's take a look at **Arguments Object** which  is an array like object but it doesn't have access to array like methods like map, reduce. Also, it's not available with arrow function. It automatically hold all the values passed to the function.
```
alodahl, line 133: Is "rest" capitalized?  I think it may make more sense to mark it as code. Basically, I would be careful how you annotate things, because this is likely the first time the learner has heard about it, and they will be trying to internalize these things. Is is a Proper Name, a `code method` to remember, a _term_? 
```suggestion
`rest` parameters allows a function to accept an indefinite numbers of arguments as an array. `rest` looks like spread, but it’s NOT spread. Before taking a look at `rest`, let's take a look at [`arguments` object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments)  which  is an array like an object but it doesn't have access to array like methods like `map()` and `reduce()`. Also, it's not available with an _arrow function_. It automatically holds all the values passed to the function.
```

javascript/newer-JS-features.md
```diff
@@ -0,0 +1,219 @@
...
+## Examples
+```javascript
+function sum() {
+    console.log(arguments) // collects all arguments passed to the function. 
+}
+sum(2,3,4,5) // Arguments(3) [2, 3, 4, callee: ƒ, Symbol(Symbol.iterator): ƒ]
+```
+It contains all values that we passed into the function by order, but we can't use array like methods like reduce. Here, let's write a function that will give sum of all the elements in the array.
```
alodahl, line 141:
```suggestion
It contains all values that we passed into the function by order, but we can't use array like methods like `reduce`. Here, let's write a function that will give sum of all the elements in the array.
```

javascript/newer-JS-features.md
```diff
@@ -0,0 +1,219 @@
...
+```javascript
+function sum() {
+    return arguments.reduce((total, acc) => total + acc);
+} 
+sum(2,3,4,5,6) // arguments.reduce is not a function
+```
+This is where Rest comes in. Rest  collects all remaining arguments or rest of the arguments and puts them into an array. The dots means gather the rest of the parameters into an array, so the rest parameters must be at the end. Rest is the opposite of Spread. In spread we're spreading things out, but in Rest we're collecting all things into a single parameter.
```
alodahl, line 148:
```suggestion
This is where `rest` comes in. `rest` collects all remaining arguments, or the _rest_ of the arguments, and puts them into an array. The dots mean  "gather the rest of the parameters into an array", so the `rest` parameters must be at the end. `rest` is the opposite of _spread_. In _spread_, we're spreading things out, but in `rest` we're collecting all things into a single parameter. 
```

javascript/newer-JS-features.md
```diff
@@ -0,0 +1,219 @@
...
+}
+console.log("The sum of array is: ", sumOfAllArray(3,4,5));  // 12
+```
+
+## Array Destructuring Assignment
+Destructuring assignment syntax allows us to unpack or extract or single out values from arrays, properties from objects into distinct variables. With array destructuring, we can copy items into variables without modifying the original array.
+
+## Examples
+### Array Destructuring
+Let's look at numbers array. variable max contains the first element of the array and secondMax contains the second element of the array and so on. Old way of doing it would be:
```
alodahl, line 164:
```suggestion
Let's look at numbers array. Variable `max` contains the first element of the array and `secondMax` contains the second element of the array and so on. Old way of doing it would be: 
```

javascript/newer-JS-features.md
```diff
@@ -0,0 +1,219 @@
...
+
+## Examples
+### Array Destructuring
+Let's look at numbers array. variable max contains the first element of the array and secondMax contains the second element of the array and so on. Old way of doing it would be: 
+```javascript
+let numbers = [6, 5, 4, 3]; 
+let max = numbers[0];
+let secondMax = numbers[1];
+```
+With ES6 new syntax, extracting or sinling out values became much easier. Instead of doing arr[0] or arr[1] we can do:
```
alodahl, line 170:
```suggestion
With ES6 new syntax, extracting or singling out values becomes much easier. Instead of doing `arr[0]` or `arr[1]` we can do:
```

javascript/newer-JS-features.md
```diff
@@ -0,0 +1,219 @@
...
+console.log(max); // Here max is holding the value of numbers[0];
+console.log(secondMax); // 5
+console.log(lowerNumbers); // We've used spread to singling out the remaining values [4,3]
+console.log(numbers); // numbers is unchanged.
+```
+### Object Destructuring
+Object destructuring is the most powerful and useful syntax as in object order doesn't matter but in array order matters. 
+```javascript
+const user = {
+    email:"renunikhilp@gmail.com",
```
alodahl, line 182: I hope this isn't real!  Maybe change these to a fictional character's info if we are going to merge this 😊

alodahl (commented):
This is really valuable, thank you!  I have some md syntax comments to help learners differentiate the terms.

alodahl:
Since I'm already being so picky, I'd suggest including mdn links at the beginning of each method for reference.

javascript/newer-JS-features.md
```diff
@@ -0,0 +1,219 @@
...
+multiply(5) // 50.
+```
+In the example above, if no value is passed for b, i.e if b is undefined, it'll get a default value of 10. This is a much nicer and shorter syntax. 
+
+## Spread Operator
+The Spread (...) operator is a useful and newer syntax for adding items to arrays, combining arrays or objects, and spreading an array out into a function’s arguments.
+## Some scenarios where Spread is useful:
+- Copying an array
+- Conncatenating or combining arrays
+- Merge two Objects.
```
alodahl, line 82:
```suggestion
- Merge two objects.
```
I don't think _objects_ is ever capitalized - I could be wrong.

javascript/newer-JS-features.md
```diff
@@ -0,0 +1,219 @@
...
+Spread in array literals creates a new array using an existing array. It basically spreads the elements from one array into a new array.
+Let's look an example -
+Here, I've 2 arrays primaryColors and secondaryColors. allColors array contain all the elemnts from primaryColors and secondaryColors. ... spread the values into seperate arguments. allColors contains copy of primaryColors and secondarycolors but the original arrays remains unchanged! 
+```javascript
+const primaryColors = ['red', 'green', 'blue'];
+const secondaryColors = ['purple','pink', 'crimson'];
+const allColors = [...primaryColors, ...secondaryColors] // (6)["red", "green", "blue", "purple", "pink", "crimson"]
+```
+### Spread with Objects
+Spread in object literal copies properties from one object into another object. In past, we've used Object.assign to copy properties from one object to another. Spread makes copying more easy!
```
alodahl, line 109: Oops, except when it's a prototype 👍🏻

javascript/newer-JS-features.md
renushinde: Thank you for your suggestions Alina and so sorry for the typos :(

renushinde (commented):
Added supplemental materials, fixed typos.

alodahl (approved):

Issue #1526, opened 2021-03-09
Update outdated usage of createReactClass in React Part 1 Slides


aedward8:
### Page where problem found?

https://github.com/Techtonica/curriculum/blob/main/react-js/react-part-1-intro.md

### Type of problem

Common examples:

-conflicting material
In the Techtonica slides on slide 12 it says "createReactClass(): this is rarely used; no need to learn this"

However, link 4 in resources "Building your first component in React.js | Pluralsight (video)" Is an 7 min example vide primarily using React.createClass.

sheriecarter:
I have updated the slides to remove "this is rarely used; no need to learn this" and instead added a bullet point that lets them know that this is not used in ES6. I think it would be potentially helpful to still keep the Pluralsight tutorial in the curriculum unless there are already other tutorials in the curriculum which go over the create-react-class module.

alodahl:
I see this is connected to #1557. I looked at the slides just now, and there is still a line that says "this is rarely used; no need to learn this", so I'm confused about what changes were made @sheriecarter .

Does this resolve the issue @aedward8 ?

aedward8:
@alodahl  I think you can resolve this issue as the larger issue of updating the curriculum from class based components to function based components is what #1557 and this issue is about

PR #1557, opened 2021-04-14
Update react-part-1-intro.md

sheriecarter:
There are a few things that will need to be updated over time. In the React slide deck, it stated that the react-create-class module did not need to be learned. That conflicted with having a tutorial use that module. Also, many of the tutorials are all done with classes instead of functions. This will be a task that I could fix in a few different steps but to start, I updated the slide and added #4 to the Common Mistakes/ Misconceptions to shed light on defining components with functions in React.

### DIFF
--- a/react-js/react-part-1-intro.md
+++ b/react-js/react-part-1-intro.md
@@ -40,7 +40,7 @@ React allows developers to create large web applications that can change data, w
 
 ### Lesson Materials
 
-You will spend the day working through these materials.
+You will spend the day working through these materials. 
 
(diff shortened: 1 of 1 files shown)


alodahl (approved):
Which slide did you update?  I assume it's within 1. [Techtonica React slides](https://docs.google.com/presentation/d/1Bswkl7e1kGVav7KFabHrBgXhRi7mlINTbinxnUcGDy8/edit?usp=sharing).

alodahl:
Addresses #1526.
@aedward8  does it solve the issue?

PR #1578, opened 2021-06-26
Add git/github configurations

xiaozhong21:
The original lesson dives directly into creating repos on Github, etc. It would be nice to have the setup and configurations of Git and Github ready before actually interacting with the terminal.


alodahl (approved):
😍

PR #1583, opened 2021-07-10
Add link + info to Temporal JS Date/Time API

picaq:
This lesson file was outdated when we covered it, so I added another option


alodahl:
@picaq what part is outdated?

picaq:
> @picaq what part is outdated?

In [momentjs.com](https://momentjs.com/), it states that
> Considering using Moment in your project?
_There may be better modern alternatives._
For more details and recommendations,
please see Project Status in the docs.

They explain [why here](https://momentjs.com/docs/#/-project-status/)
and suggest [Luxon](https://moment.github.io/luxon/#/) as an alternative that doesn’t have mutable objects.
Temporal was one that popped up in my mailing lists another JS date/time API

alodahl (approved):

alodahl:
sorry about the delay @picaq but this is an important update. thx!

PR #1595, opened 2021-08-04
Separating Preparing Paragraphs into Points

ZelmaSedano:
I think this was the author's original intention, but it was somehow altered.  Just a suggestion!


Yosolita1978:
Awesome! Thank you for this update.

PR #1596, opened 2021-08-04
Update spelling error in line 21

jenhuynh:
Change 'acccount' to 'account' in line 21 about emails


Yosolita1978:
Thank you for fixing that!

PR #1624, opened 2021-09-05
Eventonica part 2 for useReducer

dpenny:
Eventonica part 2
From this lesson, someone should be able to:
- render an event list
- render an event form to create a new event and add it to the list
- understand useReducer and how to use it in the event form


alodahl (approved):
is this ready?  🤩

Issue #1610, opened 2021-08-19
eventonica-react: help cohort add express routes to use in their create-react-app project


alodahl:
### eventonica-react: get the cohort started with express in their create-react-app project

Please add more steps after line 124 to help participants create users/delete and users/add routes to their express app. project: https://github.com/Techtonica/curriculum/blob/main/projects/eventonica-react/eventonica-express-backend.md

Feel free to proofread existing content as well.

Adding a Postgres API will be the next outline after this.  We need help with that one too! https://github.com/Techtonica/curriculum/blob/main/projects/eventonica-react/eventonica-postgres.md

Add or change the starter code in projects/eventonica-react/starter-code if necessary.
Here's an example Express project for your reference.  This is for contributors, not curriculum users! https://github.com/alodahl/eventonica-api.

See #1608 for more context and a React app example.

PR #1625, opened 2021-09-05
Eventonica Express api

alodahl:
addresses https://github.com/Techtonica/curriculum/issues/1610


alodahl:
@dpenny I have something started here but please take it over if you have time/interest.  It need to be finished by Tuesday.

PR #1689, opened 2022-06-07
Update tech-industry-tips.md

asaudu:
Here are some of the tips that stood out to me and that I was able to find. Not sure if you wanted it to have its own section, follow the format, or what, so I started with this first.


Yosolita1978:
Thanks for this!

PR #1727, opened 2022-08-10
Add Techtonica zsh lesson as a Pre-req

Yosolita1978:


cdm-ium (approved):
This is great. I wasn't aware this existed and it should help people with getting started!

PR #1748, opened 2022-09-06
Add instruction to ignore node_modules

m3ia:
After running create-react-app, it seems like we should try and gitignore node_modules right away


allejo:
Closing as this should no longer be needed as it's included in the project's `.gitignore` 

https://github.com/Techtonica/curriculum/blob/e56e074ce8b6cd4378e76561956e71daae1073c2/project-example/week-7/weather-forecast/.gitignore#L1

PR #1908, opened 2023-04-07
Adding Testing with Vite

Yosolita1978:


projects/2023TemplateWithVite/client/src/App.jsx
```diff
@@ -1,16 +1,23 @@
 import './App.css'
 import 'bootstrap/dist/css/bootstrap.min.css';
-import MyNavBar from './components/Navbar'
+import MyNavBar from './routes/Navbar'
```
allejo, line 3: I personally would leave `NavBar` in the components folder because that's what NavBar is, a shared component used on multiple pages; a route would be a page/view*.

\* At least in the traditional/beginner setup

projects/2023TemplateWithVite/client/src/App.jsx
```diff
@@ -1,16 +1,23 @@
...
+    setIsClicked(true);
+  }
+
+
   return (
     <div className="App">
-      <MyNavBar />
-      <ListStudents />
-
+      <MyNavBar handleMe={handleMe} />
```
allejo, line 19: React convention is that prop names that are callbacks start with the word "on" and then a description of what event it is. So since this callback is called whenever the account menu item is clicked, I'd rename this prop to `onAccountMenuItemClicked` or something along those lines.

And then `handleMe` can be renamed to `handleAccountMenuItemClicked` since "handle" is the conventional suffix used for implementations of callback props

projects/2023TemplateWithVite/client/src/components/contact.jsx
```diff
@@ -0,0 +1,49 @@
...
+    twitter: "yosola",
+    notes: "Some notes",
+    favorite: true,
+  };
+
+  return (
+    <div id="contact">
+      <div>
+        <img
+          key={contact.avatar}
```
allejo, line 17: A `key` prop is not necessary unless you're in a loop OR are abusing the `key` prop to explicitly trigger re-renders (which doesn't look to be the case here)

projects/2023TemplateWithVite/client/src/components/contact.jsx
```diff
@@ -0,0 +1,49 @@
...
+
+      <div>
+        <h1>
+          {contact.first || contact.last ? (
+            <>
+              {contact.first} {contact.last}
+            </>
+          ) : (
+            <i>No Name</i>
+          )}{" "}
```
allejo, line 30:
```suggestion
          )}
```

projects/2023TemplateWithVite/client/src/components/contact.jsx
```diff
@@ -0,0 +1,49 @@
...
+            </a>
+          </p>
+        )}
+
+        {contact.notes && <p>{contact.notes}</p>}
+
+        </div>
+        </div>
+        )
+    }
```
allejo, line 49:
```suggestion
      </div>
    </div>
  )
}
```

projects/2023TemplateWithVite/client/src/example.test.jsx
```diff
@@ -0,0 +1,40 @@
...
+
+test('test that MyNavBar renders', () => {
+//     const navbarElement = render(<MyNavBar/>);
+//     const navbarId = navbarElement.getByTestId('navbar');
+//     expect(navbarId).toBeDefined();
+//   });
+
+
+
+*/
```
allejo, line 40:
```suggestion
```

Looks like unused code?

projects/2023TemplateWithVite/client/src/main.jsx
```diff
@@ -1,10 +1,33 @@
...
+// /*
+// const router = createBrowserRouter(
+//   createRoutesFromElements(
+//     <Route path="/" element={<App />}>
+//       <Route path="login" element={<Contact />} />
+//       </Route>
+//       )
+//     );
+
+// */
```
allejo, line 19: Remove unused code to prevent confusion OR write an explanation on what this commented out code does

allejo (changes requested):

daaimah123 (changes requested):
this is a test comment for requesting changes

projects/2023TemplateWithVite/client/package.json
```diff
@@ -6,7 +6,8 @@
   "scripts": {
```
manojsiddoji, line 6: can i provide here

manojsiddoji (commented):
A key prop is only essential if you're working within a loop or intentionally using it to force re-renders, which doesn’t seem applicable in this situation.

daaimah123 (approved):
Updated branch.

Issue #1941, opened 2023-04-25
Incomplete resources for Code Challenges for Arrays


Yosolita1978:
_Thanks for reporting an issue with the Techtonica curriculum! Please fill in the following information to help us resolve the problem as soon as we can._

### Page where problem found?

https://github.com/Techtonica/curriculum/blob/main/code-challenges/arrays.md

### Type of problem

Common examples:

- Incomplete resources for solutions and test cases for problems about Arrays in JS.

vazquezea96:
Hello, I am interested in working on this issue.

Yosolita1978:
Perfect! I invite you to be part of our Triage team. Please accept the invite (you should also see an email) so I can assign you to this issue. Thank you for volunteering!

vazquezea96:
Awesome, thanks for the invite. So, the main goal here is to complete the test cases and solutions for the problem sets.

isadominguez314:
I would like to work on this issue

galgalinidhi:
Hello, I'm interested in working on this

YiyunBao99:
Hello, I'd like to work on this issue.

Shikha2410:
Hello, I'm interested in working on this

lyannel:
Hi there, interested in working on this issue. I've filled out the volunteer form.

Shikha2410:
#2033

PR #1989, opened 2023-08-14
#1941 incomplete resources for array code challenges

vazquezea96:
Completed the test cases and sample solution for problem one. I would like feedback on the sample solution, if it looks good, I will go ahead and do similar to the rest of the solutions.


Yosolita1978:
It's perfect! I like a lot your test cases! Thank you sooo much for doing this and helping us!

PR #2015, opened 2023-09-11
Proofreading Git Version Control

RachelGreenwood:
Note: May also be good to mention Git GUIs in here. We didn't know about them until way later, so most of us had already gotten used to the Terminal

### DIFF
--- a/git/git-version-control.md
+++ b/git/git-version-control.md
@@ -14,19 +14,19 @@
 
 ### Motivation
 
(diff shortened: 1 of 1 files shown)


allejo (approved):
> May also be good to mention Git GUIs in here. We didn't know about them until way later, so most of us had already gotten used to the Terminal

I like this idea. Create an issue with this suggestion so that any discussion on that idea can happen there 😄

PR #2021, opened 2023-09-22
Added solutions to strings problems

### LINKED ISSUE #1938: Incomplete resources for sample solutions to strings code challenges
### Page where problem found?

(https://github.com/Techtonica/curriculum/blob/main/code-challenges/string.md)

### Type of problem

 Incomplete resources for sample solutions to strings code challenges

### Suggested Solution

_Optional. If you would like to suggest a specific solution, please add it here._

archijaiswal:
Fixes #1938

### DIFF
--- a/code-challenges/string.md
+++ b/code-challenges/string.md
@@ -65,7 +65,7 @@ s2 = "eygoodb"
 Expected output: False
 
 Sample Solution:
-
+```
 function isRotation(s1, s2) {
 if (s1.length !== s2.length) {
 return false;
@@ -81,12 +81,14 @@ return str.includes(substr);
 // Implementation of the isSubstring method goes here
 // Returns true if `substr` is a substring of `str`, false otherwise
 }
-
-// Testing the isRotation function
+```
+Testing the isRotation function
+```
 const s1 = "goodbye";
 const s2 = "eygoodb";
 
-console.log(isRotation(s1, s2)); // Output: false // Fixed a mistake here
+console.log(isRotation(s1, s2)); // Output: false 
+```
 
 #### Problem 2 -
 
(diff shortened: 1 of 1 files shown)


code-challenges/string.md
```diff
@@ -65,7 +65,7 @@ s2 = "eygoodb"
 Expected output: False
 
 Sample Solution:
-
+```
```
allejo, line 68:
~~~suggestion
```javascript
~~~

You can specify what language this code is and it'll add syntax highlighting to it! Please this to all the code blocks you've touched

File: code-challenges/string.md [hunk 2]
```diff
@@ -107,8 +124,25 @@ front_times('Chocolate', 2) → 'ChoCho'
...
 front_times('Ok', 4) → 'OkOkOkOk'
 
-Needs sample solution
+Sample solution
+```
+function front_times(str, n) {
 
+  var front = str.substring(0, 3);
+  var result = "";
+  for (var i = 0; i < n; i++) {
```
allejo, line 133:
```suggestion
  const front = str.substring(0, 3);
  let result = "";
  for (let i = 0; i < n; i++) {
```

Please use `const` or `let` instead of `var`.

https://medium.com/@codingsam/awesome-javascript-no-more-var-working-title-999428999994

tl;dr:

- Use `const` when a variable won't be reassigned (i.e. you won't use an equal sign on it again)
- Use `let` when you need to reassign a value (i.e. you'll use an equal sign on it in the future)

code-challenges/string.md
```diff
@@ -176,3 +327,20 @@ Given an "out" string length 4, such as "<<>>", and a word, return a new string
 make_out_word('<<>>', 'Yay') → '<<Yay>>'
```
allejo, line 327: @Yosolita1978, for this question, is `out` always going to be of length 4? or should this solution also handle something like `##[]##`?

allejo (changes requested):

code-challenges/string.md
archijaiswal: Sure, thanks!

code-challenges/string.md
Yosolita1978: For this problem, you can assume that `out` will always be length 4.

PR #2022, opened 2023-09-22
10 array code challenge problems with examples and solutions added

isadominguez314:


isadominguez314:
i added problems 11-20 (18-20 are a little harder) to the array code practice

code-challenges/arrays.md
```diff
@@ -156,3 +156,144 @@ same_first_last([1, 2, 1]) → true
...
+
+find_odds([1,2,3,4,5,6]) → [1,3,5]
+find_odds([7,1,9,8,5,2,4,6]) → [7,1,9,5]
+find_odds([2,4,6,7]) → [7]
+
+Sample Solution: 
+
+function find_odds(arr) {
+   return arr.filter((num) => num % 2 === 1);
+}
```
allejo, line 172:
~~~suggestion
Given an array of ints, return a new array containing only the odd numbers in the same order that they appeared in the original array.

```
find_odds([1,2,3,4,5,6]) → [1,3,5]
find_odds([7,1,9,8,5,2,4,6]) → [7,1,9,5]
find_odds([2,4,6,7]) → [7]
```

Sample Solution: 

```javascript
function find_odds(arr) {
   return arr.filter((num) => num % 2 === 1);
}
```
~~~

allejo (changes requested):
I know the other examples in this document don't follow this format but since you're touching this file, could you please format the formatting of the questions/answers to this format so that it reads better when the markdown is rendered?

isadominguez314:
> I know the other examples in this document don't follow this format but since you're touching this file, could you please format the formatting of the questions/answers to this format so that it reads better when the markdown is rendered?

yes I can work on that now!

isadominguez314:
just finished the formatting changes!

PR #2197, opened 2025-01-02
Created a Topic Outline for Incremental Breakdown of Functionality in Pseudocode

themagicianking:
Issue: Create a lesson that aims to guide participants in building the skill to breakdown the logic that goes into each piece of functionality.

Feedback Request: I'd like to make sure that the lesson is not too specific to the Hoopla screen recordings and was also wondering if I should link to other screen recordings, or if that will be something provided week to week.

I have completed the following changes:

- [x] Added initial file and outline as already written in the issue.
- [x] Added motivation for learning to the lesson
- [x] Added lesson steps and worked the breakdown criteria into it instead of giving it its own section
- [x] Revised "Hoopla" to "Hoopla Digital library catalogue"
- [x] Removed redundant wording in motivation


practice/pseudocode.md
```diff
@@ -0,0 +1,27 @@
...
+- Break down logic for prospective functionality of a given problem
+- Revisit and analyze a new feature every week
+
+## Additional Requirements
+- Revisit time management slides weekly
+- Intentionally break down work implementing features and share with trio/mentor
+- Check in weekly with trio/mentor for accountability and support
+
+## Lesson
+Take a look at the following [Hoopla screen recording](https://www.dropbox.com/scl/fi/oot5swpkbiz6bvk4r0bo8/Hoopla-Digital-Screen-Recording.gif?rlkey=is70lgihzq686hcqm16lvjyu8&e=1&st=ot9djufc&dl=0). Make note of how the UI behaves when the user interacts with it. How do possible results appear when the user types? In what way might the information need to be stored so it can be displayed like this? Once the user hits enter, how are results displayed? When an item is selected, how does the UI change? What features are present on the new screen? Take note of all the features that you see in use and begin to break down how you would code that feature. For example, if you were writing pseudocode to plan out what happens when the user hits "play" on an audiobook, it might look something like this:
```
daaimah123, line 17:
```suggestion
Take a look at the following [Hoopla digital library screen recording](https://www.dropbox.com/scl/fi/oot5swpkbiz6bvk4r0bo8/Hoopla-Digital-Screen-Recording.gif?rlkey=is70lgihzq686hcqm16lvjyu8&e=1&st=ot9djufc&dl=0). Make note of how the UI behaves when the user interacts with it. How do possible results appear when the user types? In what way might the information need to be stored so it can be displayed like this? Once the user hits enter, how are results displayed? When an item is selected, how does the UI change? What features are present on the new screen? Take note of all the features that you see in use and begin to break down how you would code that feature. For example, if you were writing pseudocode to plan out what happens when the user hits "play" on an audiobook, it might look something like this:
```

daaimah123 (changes requested):

practice/pseudocode.md
```diff
@@ -0,0 +1,27 @@
+# Incremental Breakdown of Functionality in Pseudocode
+
+## Motivation
+Pseudocode is one of the most useful tools a programmer has. The ability to break down a problem into implementable steps saves you valuable time before digging into the code itself and also ensures that you fully understand the problem you're solving before you begin solving it.
```
ChasVanDav, line 4: What do you think about removing some redundant language from line 4 so that it looks like this?`Pseudocode is one of the most useful tools a programmer has. The ability to break down a problem into implementable steps saves you valuable time before digging into the code itself and also ensures that you fully understand the problem before you begin solving it.`

ChasVanDav (changes requested):

practice/pseudocode.md
themagicianking: That's much clearer, I'll change it! Thank you :)

ChasVanDav (approved):
Looks great, Thane!

practice/pseudocode.md
themagicianking: Resolved in [this](https://github.com/Techtonica/curriculum/pull/2197/commits/7a56523f93c0f2b9d882322163ea754ce9d3fa30) commit!

practice/pseudocode.md
themagicianking: Resolved in [this commit](https://github.com/Techtonica/curriculum/pull/2197/commits/ed72fbf8ae5927ce1fcbcfe475a0ad9241bc48a9)

practice/pseudocode.md
```diff
@@ -0,0 +1,27 @@
...
+- Break down logic for prospective functionality of a given problem
+- Revisit and analyze a new feature every week
+
+## Additional Requirements
+- Revisit time management slides weekly
+- Intentionally break down work implementing features and share with trio/mentor
+- Check in weekly with trio/mentor for accountability and support
+
+## Lesson
+Take a look at the following [Hoopla digital library screen recording](https://www.dropbox.com/scl/fi/oot5swpkbiz6bvk4r0bo8/Hoopla-Digital-Screen-Recording.gif?rlkey=is70lgihzq686hcqm16lvjyu8&e=1&st=ot9djufc&dl=0). Make note of how the UI behaves when the user interacts with it. How do possible results appear when the user types? In what way might the information need to be stored so it can be displayed like this? Once the user hits enter, how are results displayed? When an item is selected, how does the UI change? What features are present on the new screen? Take note of all the features that you see in use and begin to break down how you would code that feature. For example, if you were writing pseudocode to plan out what happens when the user hits "play" on an audiobook, it might look something like this:
```
MaggieFero, line 17: "Hoopla digital library" might be confusing here, because the platform is "Hoopla Digital" and you typically access it through one or more libraries, but it's not really a unified library (and some core functionality differs depending on how you access it). This might be a little clearer if you reframed as something like "screen recording of a Hoopla Digital library catalog"? For example, I have a couple libraries on my hoopla account and the UI shown here is slightly different among them, so people trying to follow along who aren't familiar might see something else and think this is outdated (versus just differently configured).

themagicianking: Thanks for this insight, Maggie! I've gone ahead and revised it in [this commit](https://github.com/Techtonica/curriculum/pull/2197/commits/a243cd93c011fbecb20efcd0644f4790d4e25d73).

ndp (approved):

daaimah123 (approved):

PR #2220, opened 2025-02-04
Update intro to React lesson

kaylahrose:


react-js/react-part-1-intro.md
```diff
@@ -22,7 +22,7 @@ React is an open-source JavaScript library that is used for building user interf
 
 React allows developers to create large web applications that can change data, without reloading the page. The main purpose of React is to be fast, scalable, and simple.
 
-[Facebook](https://www.facebook.com) created React as a framework to create reuseable components, like posts or comments, and allow the user see updates instantly. [Instagram](https://www.instagram.com/) and [Airbnb](https://www.airbnb.com/) use it too.
+[Jordan Walke](https://en.wikipedia.org/wiki/React_(software)#:~:text=React%20was%20created%20by%20Jordan,HTML%20component%20library%20for%20PHP.), a software engineer at Meta, created React as a library to create reuseable components, like posts or comments, and allow the user see updates instantly. [Instagram](https://www.instagram.com/) and [Airbnb](https://www.airbnb.com/) use it too.
```
MaggieFero, line 25: You might consider linking to the [History section](https://en.wikipedia.org/wiki/React_(software)#History) this is the first sentence of instead of the specific words, especially because the specific phrasing of a given sentence is reasonably likely to change on Wikipedia.

MaggieFero:
It looks like after your update the link still goes to text within the article? Had you meant to update to link to the history section? If so, happy to make that update. Anything with "~:text=" in the URL is going to be a link to specific text on the page, rather than a section header.

PR #2228, opened 2025-02-25
Fixed instructions on where to insert type:module

wk642:
Initial instructions suggest for `"type":"module" to be placed before scripts. However, I have found that, it does not work and actually need to be inside of scripts.

Screenshot 2025-02-24 at 7 22 34 PM


wk642:
Closing this PR, conflicting information on where type:module actually needs to be. and Also has an additional pr that is not suppose to be here

PR #2229, opened 2025-02-25
Type module location

wk642:
Fixed instructions on type:module is suppose to be inside script and not outside


daaimah123 (approved):
Great fix, go ahead and merge

PR #2792, opened 2026-09-14
Move DSA from week 11

daaimah123:
DSA in week 11, needs to be moved to optional elective lessons / independent practice


README.md
naomiquinones: Since the Week 11 section is changing, this needs to be changed as well, to "2 Full Stack Projects in 1 Week"

naomiquinones (changes requested):
One change needed; otherwise it looks good.

README.md
daaimah123: ah thanks! totally missed that — addressed in [0a3c08f](https://github.com/Techtonica/curriculum/pull/2792/commits/0a3c08fe123c6842d67f541e850f65b7e049a6a9)

naomiquinones (approved):
Thanks for updating the Table of Contents. Looks good!

